PR #1, opened 2020-06-22
Add license

betrevisan:
Add the Apache license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,203 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,203 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
kashiish, line 190: Do we need to fill this out?

betrevisan: What I got from it is that whoever uses our code will need to add their own identifying information inside the brackets but I might be wrong. I will ask in the chat.

kashiish: Yep, makes sense!

PR #2, opened 2020-06-22
Basic Registration Page

samfalberg:
These commits will create a basic registration page that allows students and tutors to sign up for our service. It will ask both students and tutors for their name and email, and tutors will be able to select the topic(s) they want to tutor in. For now, the backend will just immediately display the information that the user just inputted.

### DIFF
--- a/src/main/webapp/registration.html
+++ b/src/main/webapp/registration.html
@@ -1,0 +1,22 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>Registration Page</title>
+    <link rel="stylesheet" href="registration.css">
+    <script src="registration.js"></script>
+  </head>
+  <body>
+      <form action "/data" method="POST">
+            <p>Are you a student or a tutor?</p>
+            <input type="radio" name="role" value="student"> Student<br>
+            <input type="radio" name="role" value="tutor"> Tutor<br><br>
+            <label for="name">Name:</label>
+            <input type="text" name="name"><br>
+            <label for="email">Email:</label>
+            <input type="text" name="email"><br><br>
+
+            <input type="submit" />
+      </form>
+  </body>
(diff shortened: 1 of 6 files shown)


src/main/webapp/registration.html
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>Registration Page</title>
+    <link rel="stylesheet" href="registration.css">
+    <script src="registration.js"></script>
```
betrevisan, line 7: Kashish told me that if you put the script tag in the end of the file it will load faster. That would mean just adding this tag right before the end of the body tag. @kashiish

src/main/webapp/registration.html
```diff
@@ -0,0 +1,22 @@
...
+  </head>
+  <body>
+      <form action "/data" method="POST">
+            <p>Are you a student or a tutor?</p>
+            <input type="radio" name="role" value="student"> Student<br>
+            <input type="radio" name="role" value="tutor"> Tutor<br><br>
```
betrevisan, line 13: Can someone select both of these?

src/main/webapp/registration.html
samfalberg: Alright, fixed that

src/main/webapp/registration.html
samfalberg: Since both of their names are set to "role" it deselects the other if you try to select them both

betrevisan:
Also, make sure to add a copyright notice to the top of css and html files.

src/main/webapp/registration.html
betrevisan: Cool! It might be worth thinking how to modify this later when we allow a user to both be a student and a tutor at the same time.

PR #3, opened 2020-06-22
Update pom.xml artifact ID

kashiish:
These commits update the pom.xml artifactId to "sullivan".


kashiish:
@samfalberg I think you commented instead of approved.

PR #5, opened 2020-06-22
Basic frontend and backend for scheduling a session

betrevisan:
The commits proposed in this PR add the basic frontend and backend infrastructures for scheduling a tutoring session after a time slot has been selected. The student would have to fill out a form with further detail about the upcoming session and submit that. That would result in the user been taken to a page that confirms that the session was scheduled. I have also included the servlet for scheduling a session and the file with tests related to that. I used bootstrap for the general frontend design.


src/main/java/com/google/sps/servlets/SchedulingTest.java
```diff
@@ -0,0 +1,26 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public final class SchedulingTest {
+
+
+}
```
kashiish, line 26: New line here.

src/main/java/com/google/sps/servlets/SchedulingTest.java
```diff
@@ -0,0 +1,26 @@
+// Copyright 2019 Google LLC
```
kashiish, line 1: I think this should be in the `src/main/javatests/com/google` folder

betrevisan: Do we have that folder created yet? The tests in the walkthroughs for the start project were inside src/test/com/google/sps. If that hasn't been created yet, I can go ahead and do that.

src/main/java/com/google/sps/servlets/SchedulingTest.java
betrevisan: Just added a new one, thanks!

src/main/java/com/google/sps/servlets/SchedulingTest.java
kashiish: I don't think so. I made one in my homepage branch separately but that might cause some conflicts. Yeah, can you try creating it in master?

PR #13, opened 2020-06-23
Create a basic search results page

kashiish:
These commits create a new HTML page for search results that contain a search bar and a container for search results. I also made it so that when a user submits the search form from the homepage or the search results page, it will redirect to the search results page with data fetched from the server for the searched topic. I also added unit tests for JS and the Java servlet.


src/main/webapp/homepage.html
```diff
@@ -22,10 +22,10 @@
   </head>
   <body>
     <div id="content">
-        <div id="search-form">
+        <form id="search-form-homepage" onsubmit="return redirectToResults();">
```
betrevisan, line 25: For the future, it might be better in terms of user experience to be able to trigger redirectToResults() by just pressing enter on the keyboard. This might help: https://www.w3schools.com/howto/howto_js_trigger_button_enter.asp

src/main/webapp/search.js
```diff
@@ -12,12 +12,38 @@
...
 // limitations under the License.
 
-function doSearch() {
+/** Gets the topic the user searched for from the search box and redirects the page to the search-results page with a url that contains a query parameter for the topic. 
+    The user may already be on the search-results page. In that case, the page will reload with a different value for the topic query parameter. */
+function redirectToResults() {
+    return redirectToResultsHelper(document, window);
+}
 
+function redirectToResultsHelper(document, window) {
```
betrevisan, line 21: It might be a good idea to add a comment to this function explaining that it exists mostly just for testing purposes with Jasmine.

src/main/webapp/search.js
```diff
@@ -12,12 +12,38 @@
...
+
+    return false;
+}
+
+/** Fetches the search results for the topic that the user searched for. */
+function getSearchResults() {
+    return getSearchResultsHelper(document, window);
+}
+
+async function getSearchResultsHelper(document, window) {
```
betrevisan, line 35: Comment about the helper function for this one too

src/main/webapp/webapptests/spec/SpecSearch.js
```diff
@@ -1 +1,54 @@
...
+            spyOn(window, "fetch").and.callFake(function(servletUrl) {
+                return Promise.resolve({text: () => Promise.resolve("hello world.")});
+            });
+            await getSearchResultsHelper(mockDocument, mockWindow);
+            
+            expect(window.fetch).toHaveBeenCalledWith("/search?topic=math");
+            expect(resultContainer.innerHTML).toContain("hello world.");
+        });
+
+    });
```
betrevisan, line 51: I'm glad that you figured this one out! What was the issue?

src/main/webapp/homepage.html
kashiish: Because it's a form, it automatically has the functionality.

src/main/webapp/webapptests/spec/SpecSearch.js
kashiish: I needed to make the getSearchResultsHelper function asynchronous by using the async and await keywords!

PR #17, opened 2020-06-24
Adds cloudbuild config yaml

dieseldesai:


dieseldesai:
Force merging because the test failure is unrelated to this change and this is important to prevent future build failures from being merged accidentally

PR #19, opened 2020-06-24
Create sample data for testing

kashiish:
These commits creates a new class containing sample Tutor objects for testing.


kashiish:
@googlebot I fixed it.

PR #22, opened 2020-06-25
Display availability

betrevisan:
The commits proposed in this PR allow the user to see the availability of a tutor and select a desired available time slot for a tutoring session. This procedure is handled by the AvailabilityServlet along with the added changes to availability.java. JavaScript and Java Unit tests are included.

### DIFF
--- a/src/main/java/com/google/sps/servlets/AvailabilityServlet.java
+++ b/src/main/java/com/google/sps/servlets/AvailabilityServlet.java
@@ -31,6 +31,22 @@ import javax.servlet.*; // Can be deleted after hardcoded database is removed
 
 @WebServlet("/availability")
 public class AvailabilityServlet extends HttpServlet {
+    private static final int TIME_0800AM = TimeRange.getTimeInMinutes(8, 00);
+    private static final int TIME_0900AM = TimeRange.getTimeInMinutes(9, 00);
+    private static final int TIME_1000AM = TimeRange.getTimeInMinutes(10, 00);
+    private static final int TIME_1100AM = TimeRange.getTimeInMinutes(11, 00);
+    private static final int TIME_1200AM = TimeRange.getTimeInMinutes(12, 00);
+
+    private static final String[] SKILLS1 = new String[]{"math"};
+
+    private static final TimeRange[] AVAILABILITY1 = new TimeRange[]{TimeRange.fromStartToEnd(TIME_0800AM, TIME_0900AM), TimeRange.fromStartToEnd(TIME_1100AM, TIME_1200AM)};
+
+    private List<Tutor> hardcoded = new ArrayList<Tutor>(); 
+
+    public void init(ServletConfig servletconfig) throws ServletException { 
+        hardcoded.add(new Tutor("John", "john@gmail.com", SKILLS1, AVAILABILITY1));
+        hardcoded.add(new Tutor("Jane", "jane@gmail.com", SKILLS1, AVAILABILITY1));
+    }
 
     @Override
     public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
--- a/src/main/java/com/google/sps/servlets/AvailabilityServlet.java
+++ b/src/main/java/com/google/sps/servlets/AvailabilityServlet.java
@@ -14,11 +14,20 @@
 
 package com.google.sps.servlets;
 
+import com.google.sps.data.Tutor;
+import com.google.sps.data.TimeRange;
+import com.google.gson.Gson;
 import java.io.IOException;
+import java.util.Optional;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import javax.servlet.*; // Can be deleted after hardcoded database is removed
+
 
 @WebServlet("/availability")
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/sps/servlets/AvailabilityServlet.java
```diff
@@ -14,14 +14,39 @@
...
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import javax.servlet.*; // Can be deleted after hardcoded database is removed
+
 
 @WebServlet("/availability")
 public class AvailabilityServlet extends HttpServlet {
+    private static final int TIME_0800AM = TimeRange.getTimeInMinutes(8, 00);
```
kashiish, line 34: I think these would be good to add in the SampleData class

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
```diff
@@ -31,7 +56,27 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
 
     @Override
     public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-        response.setContentType("plain/text");
-        response.getWriter().println("To be implemented");
+        // Get the id of the tutor whose availability will be displayed.
+        String tutorID = Optional.ofNullable(request.getParameter("tutorID")).orElse(null);
```
kashiish, line 60: I don't really think it makes sense to use an Optional here if the default is null.

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
```diff
@@ -31,7 +56,27 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+        List<TimeRange> timeslots = new ArrayList<TimeRange>();
+
+        for (Tutor tutor : hardcoded) {
+            if (tutorID.equals(tutor.getEmail())) {
```
kashiish, line 65:
```suggestion
            if (tutorID.toLowerCase().equals(tutor.getEmail().toLowerCase())) {
```
Make it case insensitive?

src/main/webapp/webapptests/spec/SpecAvailability.js
```diff
@@ -0,0 +1,80 @@
...
+describe("Availability", function() {
+
+    describe("when the user requests to see a tutor's availability", function() {
+        var mockWindow = {location: {href: "availability.html?tutorID=test%40gmail.com", search: "?tutorID=test%40gmail.com"}};
+
+        it("should trigger the fetch function", function() {
+            spyOn(window, "onload").and.callFake(function() {
+                readTutorID(queryString, window);
+            });
+            spyOn(window, 'fetch').and.callThrough();
```
kashiish, line 24: I read that callThrough() executes the real implementation. Should we be doing that in tests?

src/test/java/com/google/sps/AvailabilityTest.java
```diff
@@ -14,17 +14,37 @@
...
+    private static final int TIME_0800AM = TimeRange.getTimeInMinutes(8, 00);
+    private static final int TIME_0900AM = TimeRange.getTimeInMinutes(9, 00);
+    private static final int TIME_1000AM = TimeRange.getTimeInMinutes(10, 00);
+    private static final int TIME_1100AM = TimeRange.getTimeInMinutes(11, 00);
+    private static final int TIME_1200AM = TimeRange.getTimeInMinutes(12, 00);
 
     @Test
-    public void noFailTest() {
-        Assert.assertEquals(true, true);
+    public void testConvertToJsonUsingGson() {
```
kashiish, line 35: I think you should maybe add a test for your doPost method and make sure it returns the correct response (like the correct tutor's timeslots for the id parameter).

src/main/webapp/webapptests/spec/SpecAvailability.js
betrevisan: In this case, I wanted fetch to run normally. The reason I created the spy was to be able to know in the end if had been successfully executed. I wouldn't be able to do expect(window.fetch).toHaveBeenCalled() if I hadn't spied on fetch.

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
betrevisan: I agree, I'll add this as a new commit.

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
betrevisan: That makes a lot of sense. I just removed it.

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
betrevisan: I agree. I'll submit it as new commit here.

src/main/webapp/webapptests/spec/SpecAvailability.js
kashiish: Ok, that makes sense!

src/test/java/com/google/sps/AvailabilityTest.java
betrevisan: Yes, that's important. My newest commit to this PR addresses that test.

kashiish: Awesome!

PR #23, opened 2020-06-25
Session Object Class

betrevisan:
The changes proposed in this PR create the Session object class, which is responsible for handling information regarding a scheduled tutoring session. It also adds the scheduledSessions property to the Tutor object class, which is an array of Session and keeps track of all the sessions scheduled with that specific tutor.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,12 +23,26 @@
       <version>4.0.1</version>
       <scope>provided</scope>
     </dependency>
+
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+      <version>2.8.6</version>
+    </dependency>
+
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.12</version>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>3.3.3</version>
+    </dependency>
+
   </dependencies>
 
   <build>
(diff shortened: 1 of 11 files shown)


pom.xml
```diff
@@ -23,6 +23,13 @@
       <version>4.0.1</version>
       <scope>provided</scope>
     </dependency>
+
+    <dependency>
```
kashiish, line 27: I think if you pull from master, this dependency should already be there.

src/main/java/com/google/sps/data/Session.java
```diff
@@ -0,0 +1,53 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+public final class Session {
```
dieseldesai, line 17: I'd suggest naming this something different as "Session" has a specific meaning in terms of HTTP and web which might be confusing here.

Maybe "TutorSession" or "Event"? Also please add javadocs as described here:
https://google.github.io/styleguide/javaguide.html#s7.3-javadoc-where-required

pom.xml
betrevisan: Oh, true! I'll do that :)

src/main/java/com/google/sps/data/Session.java
betrevisan: That makes a lot of sense. I pushed new commits to this PR that address that.

betrevisan:
@googlebot I fixed it.

betrevisan:
@googlebot I fixed it.

PR #28, opened 2020-06-26
Scheduling Tutoring Session

betrevisan:
The changes proposed in this PR implement the feature that allows a student to schedule a tutoring session with a tutor. It will add the tutoring session to the tutor's scheduledSessions array and delete the time slot that was previously available and is now taken by the newly scheduled session. The changes also implement the confirmation page, which will provide the user with a success message and with a list of scheduled session (this is currently the list of scheduled sessions of the tutor but after the Student object class is created, it will be the list of scheduled sessions of the student who just scheduled a session). 

The changes also make use of the SampleData class to create the hard-coded database.

Java and JavaScript test are included.


src/main/java/com/google/sps/servlets/AvailabilityServlet.java
```diff
@@ -62,7 +52,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
 
         List<TimeRange> timeslots = new ArrayList<TimeRange>();
 
-        for (Tutor tutor : hardcoded) {
+        for (Tutor tutor : hardcoded.getSampleTutors()) {
```
kashiish, line 55:
```suggestion
        for (Tutor tutor : SampleData.getSampleTutors()) {
```
Since all of the methods and data are static, I don't think you have to create a new instance of SampleData.

src/main/java/com/google/sps/servlets/ConfirmationServlet.java
```diff
@@ -0,0 +1,73 @@
...
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.*; // Can be deleted after hardcoded database is removed
+
+
+@WebServlet("/confirmation")
+public class ConfirmationServlet extends HttpServlet {
+    SampleData hardcoded;
+
+    public void init(ServletConfig servletconfig) throws ServletException { 
+        hardcoded = new SampleData();
```
kashiish, line 39: Same thing here. You can just do `SampleData.<whatever method>`

src/test/java/com/google/sps/SchedulingTest.java
```diff
@@ -14,17 +14,58 @@
...
+        verify(request, atLeast(1)).getParameter("tutorID");
+        verify(request, atLeast(1)).getParameter("start");
+        verify(request, atLeast(1)).getParameter("end");
+        verify(request, atLeast(1)).getParameter("studentEmail");
+        verify(request, atLeast(1)).getParameter("subtopics");
+        verify(request, atLeast(1)).getParameter("questions");
+
+        writer.flush();
+        // Tutoring session should have been scheduled
+        Assert.assertTrue(stringWriter.toString().contains("{\"studentEmail\":\"test@gmail.com\",\"tutorEmail\":\"btrevisan@google.com\",\"subtopics\":\"algebra\",\"questions\":\"How does it work?\",\"timeslot\":{\"start\":480,\"duration\":120,\"end\":600}}"));
```
kashiish, line 66: I think Anand said we can use Gson here so that we don't have to write the JSON strings ourselves.

src/main/java/com/google/sps/servlets/ConfirmationServlet.java
```diff
@@ -0,0 +1,73 @@
...
+
+        List<TutorSession> scheduledSessions = new ArrayList<TutorSession>();
+
+        for (Tutor tutor : hardcoded.getSampleTutors()) {
+            if (tutorID.toLowerCase().equals(tutor.getEmail().toLowerCase())) {
+                scheduledSessions = Arrays.asList(tutor.getScheduledSessions());
+            }
+        }
+
+        String json = convertToJsonUsingGson(scheduledSessions);
```
samfalberg, line 61: Is having a separate function to convert to Gson easier? I just write "String json = new Gson().toJson(object)" and I think that's easier to understand.

src/main/java/com/google/sps/servlets/SchedulingServlet.java
```diff
@@ -31,7 +45,33 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+        TutorSession tutoringSession = new TutorSession(studentEmail, tutorID, subtopics, questions, timeslot);
+
+        // Update scheduledSessions
+        hardcoded.addToScheduledSessionsByEmail(tutorID, tutoringSession);
+
+        // Remove available timeslot
+        hardcoded.deleteAvailabilityByTimeRange(tutorID, timeslot);
+
+        String json = convertToJsonUsingGson(hardcoded.getSampleTutors());
```
samfalberg, line 65: (same question about using a separate function for gson)

src/main/java/com/google/sps/servlets/SchedulingServlet.java
```diff
@@ -31,7 +45,33 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     @Override
     public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-        response.setContentType("plain/text");
-        response.getWriter().println("To be implemented");
+        String tutorID = request.getParameter("tutorID");
+        String start = request.getParameter("start");
+        String end = request.getParameter("end");
+        String studentEmail = request.getParameter("studentEmail");
+        String subtopics = request.getParameter("subtopics");
+        String questions = request.getParameter("questions");
```
samfalberg, line 53: Maybe use Optionals to make default values for these Strings?

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
betrevisan: Oh that makes it so much better, thanks!

src/main/java/com/google/sps/servlets/ConfirmationServlet.java
betrevisan: Thanks!

src/test/java/com/google/sps/SchedulingTest.java
betrevisan: Cool! I just committed those changes.

src/main/java/com/google/sps/servlets/ConfirmationServlet.java
betrevisan: I like this a lot better, thanks!

src/main/java/com/google/sps/servlets/SchedulingServlet.java
betrevisan: These all already have null as their default values, which I think should be the default if a user has not filled out those fields.

PR #29, opened 2020-06-26
Add search functionality

kashiish:
These commits implement the doGet method in the Search servlet so that it returns a list of Tutors for a searched topic, along with a few tests for the servlet. I also created search result elements for each result and displayed them in the result-container div. I updated the test for the result container and added tests for creating the result elements. Lastly, I added some styling using Bootstrap and CSS.

### DIFF
--- a/src/main/webapp/homepage.html
+++ b/src/main/webapp/homepage.html
@@ -1,11 +1,8 @@
- <!-- Copyright 2019 Google LLC
-
+<!-- Copyright 2019 Google LLC
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
  You may obtain a copy of the License at
-
      https://www.apache.org/licenses/LICENSE-2.0
-
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -23,10 +20,10 @@ See the License for the specific language governing permissions and
     <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
   </head>
   <body>
-    <div id="content">
-        <form id="search-form-homepage" onsubmit="return redirectToResults();">
-            <input type="text" id="topic-search-box" placeholder="What do you need help with today?">
-            <button type="submit">Go</button>
+    <div id="content" class="container-fluid">
+        <form id="search-form-homepage" class="center-block" onsubmit="return redirectToResults();">
+            <div class="form-group-lg"><input class="form-control search-box" type="text" id="search-box-homepage" required="" aria-required=true placeholder="What do you need help with today?"></div>
+            <button class="btn btn-lg" type="submit" id="search-submit-homepage">Go</button>
         </form>
     </div>
     <script src="search.js"></script>
--- a/src/main/webapp/homepage.html
+++ b/src/main/webapp/homepage.html
@@ -18,6 +18,8 @@ See the License for the specific language governing permissions and
     <meta charset="UTF-8">
     <title>Sullivan</title>
     <link rel="stylesheet" type="text/css" href="homepage.css">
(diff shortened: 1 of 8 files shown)


src/main/webapp/homepage.html
```diff
@@ -18,13 +15,15 @@
...
     <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
   </head>
   <body>
-    <div id="content">
-        <form id="search-form-homepage" onsubmit="return redirectToResults();">
-            <input type="text" id="topic-search-box" placeholder="What do you need help with today?">
-            <button type="submit">Go</button>
+    <div id="content" class="container-fluid">
+        <form id="search-form-homepage" class="center-block" onsubmit="return redirectToResults();">
+            <div class="form-group-lg"><input class="form-control search-box" type="text" id="search-box-homepage" required="" aria-required=true placeholder="What do you need help with today?"></div>
```
betrevisan, line 25: It's great that you also used some Bootstrap classes! :)

src/main/webapp/search-results.html
```diff
@@ -18,16 +15,18 @@
...
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
     <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
   </head>
   <body onload="getSearchResults()">
-    <div id="content">
-        <form id="search-form" onsubmit="return redirectToResults();">
-            <input type="text" id="topic-search-box" placeholder="What do you need help with today?">
-            <button type="submit">Go</button>
+    <div id="content" class="container-fluid" style="text-align: center;">
+        <form id="search-form-results" onsubmit="return redirectToResults();">
```
betrevisan, line 24: I think that you could just say onsubmit="redirectToResults();" here. Same thing would go for search.html.

src/main/webapp/search.js
```diff
@@ -43,9 +43,54 @@ async function getSearchResultsHelper(document, window) {
...
             var resultContainer = document.getElementById("result-container");
-            resultContainer.innerHTML = results;
+
+            var numSearchResults = document.createElement("h4");
+            numSearchResults.id = "num-search-results";
+
+            resultContainer.appendChild(numSearchResults);
+
+            //if there was an error reported by the servlet, display the error message
+            if(results.error) {
```
betrevisan, line 55: I didn't know about this error method. Looks helpful! I'll implement that in my js function too then :)

src/test/java/com/google/sps/SearchTest.java
```diff
@@ -14,17 +14,118 @@
...
+        when(response.getWriter()).thenReturn(writer);
+
+        //create the hard coded data
+        servlet.doGet(request, response);
+
+        //verify that getParameter was called
+        verify(request, atLeast(1)).getParameter("topic"); 
+        writer.flush(); // it may not have been flushed yet...
+        List<Tutor> expectedTutorList = Arrays.asList(new Tutor("Kashish Arora", "kashisharora@google.com", new String[]{"Math", "History"}, new TimeRange[]{TimeRange.fromStartToEnd(720, 780), TimeRange.fromStartToEnd(900,1020)}, new TutorSession[]{}));
+        String expected = convertListToJson(expectedTutorList);
```
betrevisan, line 61: Sam suggested getting rid of the gson helper functions and simply writing String expected = new Gson().toJson(object); and I thought that made things cleaner

src/main/webapp/search.js
kashiish: I just return a JSON string with an error attribute in my servlet if the user's search was invalid. I don't believe it has that attribute automatically.

betrevisan: Oh I see it now! Good idea

src/main/webapp/search-results.html
kashiish: I think I need that the `return` because the form will not stop submission until it finds a return statement (in which case it won't redirect the page).

src/test/java/com/google/sps/SearchTest.java
kashiish: That works better, thanks!

PR #33, opened 2020-06-30
Add date to TimeRange

betrevisan:
The commits proposed in this PR add a date property of the Calendar class to the TimeRange object class. This will be used to provide the user with enough information about the available timeslots and also to be able to sort tutoring sessions based on sessions that have happened already and the ones that are still yet to happen. Other changes include updating calls to TimeRange functions in the java servlets and in the js function, alongside their test files.


src/main/java/com/google/sps/data/SampleData.java
```diff
@@ -29,12 +30,15 @@
...
-        new Tutor("Sam Falberg", "sfalberg@google.com", new String[]{"Geology", "English"}, new TimeRange[]{TimeRange.fromStartToEnd(TIME_1000AM, TIME_1200AM), TimeRange.fromStartToEnd(TIME_0100PM,TIME_0200PM)}, new TutorSession[]{}),
-        new Tutor("Anand Desai", "thegoogler@google.com", new String[]{"Finance", "Chemistry"}, new TimeRange[]{TimeRange.fromStartToEnd(TIME_1000AM, TIME_1200AM), TimeRange.fromStartToEnd(TIME_0100PM,TIME_0200PM)},  new TutorSession[]{}),
-        new Tutor("Elian Dumitru", "elian@google.com", new String[]{"Geology", "Math"}, new TimeRange[]{TimeRange.fromStartToEnd(TIME_1000AM, TIME_1200AM), TimeRange.fromStartToEnd(TIME_0100PM,TIME_0200PM)},  new TutorSession[]{})
+        new Tutor("Kashish Arora", "kashisharora@google.com", new String[]{"Math", "History"}, new TimeRange[]{TimeRange.fromStartToEnd(TIME_1200AM, TIME_0100PM, MAY182020), TimeRange.fromStartToEnd(TIME_0300PM,TIME_0500PM, JUNE102020)}, new TutorSession[]{}),
```
dieseldesai, line 37: Can you add some more newlines here for readability? Usually we try to limit line lengths to 100 chars in Java

src/test/java/com/google/sps/SchedulingTest.java
```diff
@@ -58,12 +73,17 @@ public void testDoPost() throws Exception {
         verify(request, atLeast(1)).getParameter("tutorID");
         verify(request, atLeast(1)).getParameter("start");
         verify(request, atLeast(1)).getParameter("end");
+        verify(request, atLeast(1)).getParameter("year");
```
dieseldesai, line 76: should these be checking exactly 1 instead of at least?

src/main/java/com/google/sps/data/SampleData.java
betrevisan: Yes, I just added a new commit to this PR that fixes that in this file and in some other files that had the same issue.

src/test/java/com/google/sps/SchedulingTest.java
betrevisan: That would make it more accurate, yes. I added those changes in a new commit. Thanks!

PR #36, opened 2020-06-30
Feature Integration

samfalberg:
These commits will integrate the many features we've been working on. This will make it so the user can navigate the site easily without having to manually change the URL. Commits will include adding things like a homepage and login/logout link to every page. When a user logs in or out, the webapp should navigate them back to the page they were previously on.

### DIFF
--- a/src/main/java/com/google/sps/servlets/LoginStatusServlet.java
+++ b/src/main/java/com/google/sps/servlets/LoginStatusServlet.java
@@ -21,6 +21,7 @@ import com.google.gson.Gson;
 import com.google.sps.data.LoginStatus;
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.net.URL;
 import java.util.Iterator;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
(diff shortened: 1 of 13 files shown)


samfalberg:
A homepage link will be in the top left corner of every page. Eventually we'll replace this with the Sullivan logo, which will also direct you to the homepage when clicked.

Screenshot 2020-07-01 at 11 26 15 AM

samfalberg:
A login/logout link will be displayed at the top right of every page. If the user is logged in, it will show the logout link, which will simply logout the user but keep them at the page they are currently on. If the user is logged out, it will show the login link, which will redirect the user to the registration page.

Screenshot 2020-07-01 at 12 41 16 PM

src/main/webapp/webapptests/spec/SpecRegistration.js
```diff
@@ -65,69 +65,95 @@ describe("When displaying registration form", function() {
...
-        fetchLoginStatusHelper(document, mockLoginStatus);
+        // it("sets login link correctly", function() {
+        //     mockLoginStatus = {isLoggedIn:false, needsToRegister:false, url:'/_ah/login?continue=%2Fregistration.html'};
+        //     fetchLoginStatusHelper(document, mockLoginStatus);
 
-        expect(document.getElementById('login-form').style.display).toBe('block');
-        expect(document.getElementById('logout-form').style.display).toBe('none');
-        expect(document.getElementById('registration-form').style.display).toBe('none');
+        //     expect(document.getElementById('login-url').href).toBe('http://localhost:8080/_ah/login?continue=%2Fregistration.html');
+        // });
```
kashiish, line 104: Delete?

src/main/java/com/google/sps/servlets/LoginStatusServlet.java
```diff
@@ -45,18 +47,20 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-        setRegistered(response, name);
+        setRegistered(request, response, name);
     }
 
     /** Determines if a logged in user needs to register, created for testing */
-    public void setRegistered(HttpServletResponse response, String name) throws IOException {
+    public void setRegistered(HttpServletRequest request, HttpServletResponse response, String name) throws IOException {
         LoginStatus loginStatus;
+        // Gets the referring URL of the user so we can return them to the same page when they logout
+        String referrer = request.getHeader("referer");     //"referer" is intentionally misspelled
```
betrevisan, line 58: I couldn't really find where header is being set. Could you explain really briefly how it worls?

samfalberg: I'm not setting the header anywhere. getHeader("referer") just returns the address of whatever page the user was at before clicking the logout link.
For further information: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer

PR #38, opened 2020-07-01
Rating

betrevisan:
The commits proposed in this PR implement the rating a tutor feature. Changes include creating a page to see a tutoring session history, creating a history servlet to handle displaying a student's tutoring session history, and creating a rating servlet to rate a tutor.

### DIFF
--- a/src/main/java/com/google/sps/data/SampleData.java
+++ b/src/main/java/com/google/sps/data/SampleData.java
@@ -40,39 +40,39 @@ public final class SampleData {
                                                         .build();
 
     private static ArrayList<Tutor> tutors = new ArrayList<Tutor> (Arrays.asList(
-        new Tutor("Kashish Arora", "kashisharora@google.com", new String[]{"Math", "History"},
-                new TimeRange[]{TimeRange.fromStartToEnd(TIME_1200AM, TIME_0100PM, MAY182020),
-                            TimeRange.fromStartToEnd(TIME_0300PM,TIME_0500PM, JUNE102020)},
-                new TutorSession[]{}),
-        new Tutor("Bernardo Eilert Trevisan", "btrevisan@google.com", new String[]{"English", "Physics"},
-                new TimeRange[]{TimeRange.fromStartToEnd(TIME_0800AM, TIME_1000AM, MAY182020),
-                             TimeRange.fromStartToEnd(TIME_1100AM,TIME_0100PM, JUNE102020)},
-                new TutorSession[]{}),
-        new Tutor("Sam Falberg", "sfalberg@google.com", new String[]{"Geology", "English"},
-                new TimeRange[]{TimeRange.fromStartToEnd(TIME_1000AM, TIME_1200AM, MAY182020),
-                            TimeRange.fromStartToEnd(TIME_0100PM,TIME_0200PM, JUNE102020)},
-                new TutorSession[]{}),
-        new Tutor("Anand Desai", "thegoogler@google.com", new String[]{"Finance", "Chemistry"},
-                new TimeRange[]{TimeRange.fromStartToEnd(TIME_1000AM, TIME_1200AM, MAY182020),
-                            TimeRange.fromStartToEnd(TIME_0100PM,TIME_0200PM, JUNE102020)}, 
-                new TutorSession[]{}),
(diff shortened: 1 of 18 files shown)


src/main/java/com/google/sps/data/SampleData.java
```diff
@@ -172,4 +172,16 @@ public static void addToStudentScheduledSessionsByEmail(String email, TutorSessi
         students.add(student);
     }
 
+    public static void rateTutorByEmail(String tutorEmail, String studentEmail, int newRating) {
+        Tutor tutor = getTutorByEmail(tutorEmail);
+        Student student = getStudentByEmail(studentEmail);
+
+        tutors.remove(tutor);
+        tutor.rateTutor(newRating);
+        tutors.add(tutor);
```
kashiish, line 181: Do you have to remove the tutor and add again? The tutor object should get updated in the `tutors` list when you call `rateTutor` on the tutor object.

src/main/java/com/google/sps/data/SampleData.java
```diff
@@ -172,4 +172,16 @@ public static void addToStudentScheduledSessionsByEmail(String email, TutorSessi
...
+
+        students.remove(student);
+        student.markTutoringSessionAsRatedByTutorEmail(tutorEmail, newRating);
+        students.add(student);
```
kashiish, line 185: Same thing here. I don't think you have to remove and add the student again because you're just updating the student object.

src/main/java/com/google/sps/data/Student.java
```diff
@@ -61,4 +61,17 @@ public void addToScheduledSessions(TutorSession tutoringSession) {
         this.scheduledSessions = scheduledSessionsList.toArray(new TutorSession[0]);
     }
 
+    /** Marks the tutoring session that has the given tutor's email as rated.*/
+    public void markTutoringSessionAsRatedByTutorEmail(String tutorEmail, int newRating) {
+        List<TutorSession> scheduledSessionsList = new LinkedList<TutorSession>(Arrays.asList(this.scheduledSessions));
```
kashiish, line 66: I think it might be better if we made the learning and scheduledSessions array an ArrayList instead so we don't have to convert every time. I think this would also help with datastore because we can store Lists but not arrays in the entities.

src/main/java/com/google/sps/servlets/RatingServlet.java
```diff
@@ -0,0 +1,57 @@
...
+        String tutorEmail = request.getParameter("tutorEmail");
+        String studentEmail = request.getParameter("studentEmail");
+        int rating = Integer.parseInt(request.getParameter("rating"));
+
+        // Update tutor's rating
+        SampleData.rateTutorByEmail(tutorEmail, studentEmail, rating);
+
+        String jsonTutors = new Gson().toJson(SampleData.getSampleTutors());
+        String jsonStudents = new Gson().toJson(SampleData.getSampleStudents());
+        String json = jsonTutors + ", " + jsonStudents;
```
kashiish, line 51: I wonder if this would work so you can send an array instead of just comma separated strings:
```suggestion
        String json = new Gson().toJson(new String[]{jsonTutors, jsonStudents}); 
```

src/main/java/com/google/sps/data/Student.java
kashiish: I'll actually make a PR for this and change the arrays to ArrayLists.

src/main/java/com/google/sps/data/SampleData.java
betrevisan: The reason why I was doing this was because of the limitations of the Arrays but now that they will be ArrayLists I think this should be much easier. I will update the functions after Arrays lists has been merged then.

src/main/java/com/google/sps/data/Student.java
betrevisan: That would be great! Thank you for doing that

src/main/java/com/google/sps/servlets/RatingServlet.java
betrevisan: That worked! Thanks :)

src/main/java/com/google/sps/data/SampleData.java
betrevisan: Just committed changes that update these functions

PR #44, opened 2020-07-06
Add time as a filter in history

betrevisan:
The commits proposed in this PR will filter the the tutoring sessions for the history page so that only previous tutoring sessions show up. The converse applies for the confirmation page; only upcoming sessions will show up in the confirmation page.

### DIFF
--- a/src/main/java/com/google/sps/servlets/HistoryServlet.java
+++ b/src/main/java/com/google/sps/servlets/HistoryServlet.java
@@ -25,6 +25,8 @@ import java.util.Optional;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Date;
+import java.util.Calendar;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/sps/servlets/HistoryServlet.java
```diff
@@ -53,9 +55,35 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+            int start = session.getTimeslot().getStart();
+            int hour = start / 60;
+            int minute = start % 60;
+
+            sessionCalendar.set(Calendar.HOUR_OF_DAY, hour);
+            sessionCalendar.set(Calendar.MINUTE, minute);
+
+            Date sessionDate = sessionCalendar.getTime();
+            int comparison = currentDate.compareTo(sessionDate);
+            if (comparison == 0 || comparison == 1) {
```
samfalberg, line 82: Since this and the upcoming sessions logic add sessions when the current date is equal to the session date, couldn't a session appear both on the history and upcoming sessions page?

betrevisan: Yes, this would be the case if a session time and date equal the current time and date. I think it would be fine for it to be displayed on both upcoming and history since it would mean that the session is currently happening. IN addition, this would only happen for one minute because after that the current time would not be equal to the session time and that session would, thus, only show up on the history page.

samfalberg: Oh I see. Thanks for clarifying

PR #46, opened 2020-07-07
Integrate Datastore for search functionality

kashiish:
These commits create a new interface called SearchDatastoreService.java that has a method to get a list of tutors with a specified topic. I created two new classes that implement this interface, RealSearchDatastore which uses the actual App Engine Datastore and MockSearchDatastore which uses the SampleData class.


src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -32,6 +34,22 @@
 /** Servlet that returns a list of tutors for a searched topic. */
 @WebServlet("/search")
 public class SearchServlet extends HttpServlet {
+    private SearchDatastoreService datastore;
+
+    public SearchServlet() {}
```
betrevisan, line 39: Why is this one empty? Shouldn't just the one below be kept instead?

src/main/java/com/google/sps/utilities/RealSearchDatastore.java
```diff
@@ -0,0 +1,159 @@
...
+import com.google.appengine.api.datastore.TransactionOptions;
+import java.lang.String;
+import java.util.Collection;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Calendar;
+import com.google.gson.Gson;
+
+/** Accesses Datastore to implement the SearchDatastoreService methods. */
+public final class RealSearchDatastore implements SearchDatastoreService {
```
betrevisan, line 42: This whole file looks great! Good job at making it thread safe too

src/main/java/com/google/sps/servlets/SearchServlet.java
kashiish: Good question! I should probably add a comment to clarify this. Because I specified a constructor with a parameter (the testing one), the default empty constructor does not work anymore so I have to explicitly create it. We need the default one for deployment because the servlet is created without parameters.

betrevisan: Oh I see! Sounds good :)

PR #47, opened 2020-07-07
User Profiles

samfalberg:
These commits will add extra fields to the registration page such as profile picture and user biography. Users will be allowed to navigate to their own profile via a button at the top right of every page.


samfalberg:
Here's what the homepage looks like with the "My Account" button. When clicked, it will navigate the user to "profile.html/userId". So, if the user's ID is 12345, the URL will become "profile.html/12345".

Screenshot 2020-07-07 at 5 32 15 PM

src/main/java/com/google/sps/servlets/RegistrationServlet.java
```diff
@@ -46,6 +46,9 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     String lastName = Optional.ofNullable(request.getParameter("last-name"))
             .orElseThrow(() -> new IllegalArgumentException("Must fill out last name"));
     String fullName = firstName + " " + lastName;
+    String bio = Optional.ofNullable(request.getParameter("bio")).orElse(null);
```
kashiish, line 49: Instead of null, should this be an empty string? If we want to keep it null, I don't think we need an Optional here.

src/main/java/com/google/sps/servlets/ProfileServlet.java
```diff
@@ -0,0 +1,121 @@
...
+        String userId = Optional.ofNullable(request.getParameter("user-id")).orElse(null);
+        
+        // Find out whether the user is a student or a tutor
+        Query query = new Query("User").setFilter(new Query.FilterPredicate("userId", Query.FilterOperator.EQUAL, userId));
+        PreparedQuery results = datastore.prepare(query);
+        Entity entity = results.asSingleEntity();
+        String role = (String) entity.getProperty("role"); 
+
+        // User is a student, get their info
+        if (role.toLowerCase().equals("student")) {
```
kashiish, line 81: We definitely don't have to do this now, but it might be easier if we had Student and Tutor as a subclass of User since they have some of the same basic properties for profiles. What do you think?

src/main/webapp/profile.js
```diff
@@ -0,0 +1,62 @@
...
+// limitations under the License.
+
+// Fetch info about the spceific user logged in 
+function displayProfile() {
+    // Get the userID parameter from the URL
+    var params = new URL(location.href).searchParams;
+    var userID = params.get('userID');
+    console.log(userID);
+
+    fetch('/profile?user-id='+userID).then(response => response.json()).then((user) => {
```
kashiish, line 22: What happens if there is no user logged in? Is the My Account link still visible from other page?

src/main/java/com/google/sps/servlets/ProfileServlet.java
```diff
@@ -0,0 +1,121 @@
...
+
+            String json = new Gson().toJson(tutor);
+            response.getWriter().println(json);
+        }
+    }
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+
+    }
```
betrevisan, line 120: Would this method be used to edit/update a user's bio and profile picture or do you envision that being elsewhere?

samfalberg: I was planning to have the editing option here

src/main/java/com/google/sps/servlets/ProfileServlet.java
samfalberg: Yes that would be easier. Then we wouldn't have to repeat code for things like name and email.

src/main/webapp/profile.js
samfalberg: In the displayLoginLogoutLink function in registration.js I have it so the button will only display when the user is logged in. I just included this in that function since I thought it would be easier to have all login/logout display logic together.

src/main/java/com/google/sps/servlets/RegistrationServlet.java
samfalberg: Alright, I changed it

PR #48, opened 2020-07-07
Integrate Datastore to get tutor availability

kashiish:
These commits create a new interface called AvailabilityDatastoreService.java that has three methods: get a list of time ranges for a specified tutor, add a new time range to the tutor's availability, and delete a time range from the tutor's availability. I created two new classes that implement this interface, RealAvailabilityDatastore which uses the actual App Engine Datastore and MockAvailabilityDatastore which uses the SampleData class.


src/main/java/com/google/sps/servlets/AvailabilityServlet.java
```diff
@@ -28,10 +28,27 @@
...
+import com.google.sps.utilities.RealAvailabilityDatastore;
+import com.google.sps.utilities.MockAvailabilityDatastore;
+import com.google.sps.utilities.AvailabilityDatastoreService;
 
 @WebServlet("/availability")
 public class AvailabilityServlet extends HttpServlet {
 
+    private AvailabilityDatastoreService datastore;
+
+    public AvailabilityServlet(){}
```
betrevisan, line 40: I guess a similar comment describing the use of this would be helpful like in the search servlet.

src/main/java/com/google/sps/utilities/AvailabilityDatastoreService.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.List;
+
+/** Interface for accessing datastore to get tutor availability . */
+public interface AvailabilityDatastoreService {
+
+    /**
+    * Gets the availability of a tutor with the given email.
+    * @return List<TimeRange>
+    */
+    public List<TimeRange> getAvailabilityForTutor(String email);
```
betrevisan, line 28: Would this be where we would add functions like addAvailability or removeAvailability too?

kashiish: Yeah, I'll add those methods!

src/main/java/com/google/sps/data/Tutor.java
```diff
@@ -65,6 +65,10 @@ public Float getRating() {
         return (float) this.ratingSum / this.ratingCount;
     }
 
+    public void addAvailabilityByTimeRange(TimeRange timeslot) {
+        this.availability.add(timeslot);
+    }
+
```
betrevisan, line 71: I think I just wrote the exact function in my "Manage Tutor's Availability" PR but that's totally fine.

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
```diff
@@ -28,10 +28,32 @@
...
+    public AvailabilityServlet(boolean test) {
+        if(test) {
+            datastore = new MockAvailabilityDatastore();
+        }
+    }
+
+    public void init() {
+        datastore = new RealAvailabilityDatastore();
+    }
+
```
betrevisan, line 56: Why do we need to initialize a new one here?

kashiish: The variable `datastore` is not initialized before this so it gets initialized in the `init` method with the real datastore when the servlet is created. The tests don't call `init`, so `datastore` gets initialized in the test constructor with the mock implementation.

src/main/java/com/google/sps/data/Tutor.java
kashiish: Oh yeah, I noticed that. I'll delete this before merging.

src/main/java/com/google/sps/servlets/AvailabilityServlet.java
```diff
@@ -28,10 +28,32 @@
...
 
+    private AvailabilityDatastoreService datastore;
+
+    /**
+    * Because we created a constructor with a parameter (the testing one), the default empty constructor does not work anymore so we have to explicitly create it. 
+    * We need the default one for deployment because the servlet is created without parameters.
+    */
+    public AvailabilityServlet(){}
+
+    public AvailabilityServlet(boolean test) {
```
samfalberg, line 47: Would we need to add something like this to every servlet that gets the availability of a tutor so that we can test properly?

kashiish: Yes, we would need to add a new constructor like this to every servlet that uses a datastore service class (not just getting availability).

src/main/java/com/google/sps/data/Tutor.java
kashiish: Just realized I can't delete this because it messes up the tests. We might just have to take care of some merge conflicts.

PR #49, opened 2020-07-07
Store TimeRange and TutorSession entities with a tutor id

kashiish:
These commits remove the availability and tutor session fields from the tutor and student entities. Instead of storing a list of ids inside of the entity, a tutor id will be stored with the TimeRange and TutorSession entities. This makes it easier to add and delete time ranges and sessions because we won't have to keep updating the lists inside of the Tutor entity every time. Whenever we need to get the availability of a tutor, we can easily filter out TimeRange entities based on their tutor id or whenever we need to get a list of tutoring sessions, we can filter out TutorSession entities based on tutor or student id.

### DIFF
--- a/src/main/java/com/google/sps/utilities/RealSearchDatastore.java
+++ b/src/main/java/com/google/sps/utilities/RealSearchDatastore.java
@@ -66,7 +66,7 @@ public final class RealSearchDatastore implements SearchDatastoreService {
                 String name = (String) tutorEntity.getProperty("name");
                 String email = (String) tutorEntity.getProperty("email");
                 ArrayList skills = (ArrayList) tutorEntity.getProperty("topics");
-                ArrayList<TimeRange> availability = getTimeRanges(datastore, id);
+                ArrayList<TimeRange> availability = getTimeRanges(datastore, txn, tutorEntity.getKey());
                 ArrayList<TutorSession> scheduledSessions = getScheduledSessions(datastore, txn, id);
 
                 Tutor tutor = new Tutor(name, email, skills, availability, scheduledSessions);
@@ -126,14 +126,12 @@ public final class RealSearchDatastore implements SearchDatastoreService {
     * Gets all the time range entities corresponding that have the given tutorId.
     * @return ArrayList<TimeRange>
     */
-    private ArrayList<TimeRange> getTimeRanges(DatastoreService datastore, long tutorId) {
+    private ArrayList<TimeRange> getTimeRanges(DatastoreService datastore, Transaction txn, Key tutorKey) {
         ArrayList<TimeRange> availability = new ArrayList<TimeRange>();
         
-        //get all time ranges with tutor id
-        Filter filter = new FilterPredicate("tutorId", FilterOperator.EQUAL, tutorId);
-        Query query = new Query("TimeRange").setFilter(filter);
-
-        PreparedQuery timeRanges = datastore.prepare(query);
+        //Use ancestor query to get all time ranges that belong to the tutor
+        Query query = new Query("TimeRange", tutorKey);
+        PreparedQuery timeRanges = datastore.prepare(txn, query);
 
         for(Entity time : timeRanges.asIterable()) {
             availability.add(createTimeRange(time));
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/utilities/RealSearchDatastore.java
```diff
@@ -61,12 +61,13 @@
...
             for (Entity tutorEntity : tutorResults.asIterable()) {
-
+                
+                long id = (long) tutorEntity.getKey().getId();
                 String name = (String) tutorEntity.getProperty("name");
                 String email = (String) tutorEntity.getProperty("email");
                 ArrayList skills = (ArrayList) tutorEntity.getProperty("topics");
-                ArrayList<TimeRange> availability = getTimeRanges(datastore, txn, (ArrayList) tutorEntity.getProperty("availability"));
-                ArrayList<TutorSession> scheduledSessions = getScheduledSessions(datastore, txn, (ArrayList) tutorEntity.getProperty("scheduledSessions"));
+                ArrayList<TimeRange> availability = getTimeRanges(datastore, id);
```
betrevisan, line 69: Shouldn't we still use getTimeRanges(datastore, txn, id) here?

src/main/java/com/google/sps/utilities/RealSearchDatastore.java
```diff
@@ -86,69 +87,66 @@
...
         return sessions;
     }
 
     /**
-    * Gets all the time range entities corresponding to the ids in rangeIds and creates TimeRange objects.
+    * Gets all the time range entities corresponding that have the given tutorId.
     * @return ArrayList<TimeRange>
     */
-    private ArrayList<TimeRange> getTimeRanges(DatastoreService datastore, Transaction txn, List<Long> rangeIds) {
+    private ArrayList<TimeRange> getTimeRanges(DatastoreService datastore, long tutorId) {
```
betrevisan, line 129: Why aren't we using the transaction here anymore?

src/main/java/com/google/sps/utilities/RealSearchDatastore.java
kashiish: I couldn't use transactions here because they don't work with the type of query I was using. They only work with ancestor queries, so the TimeRange entities need to have the tutor entity as an ancestor. Here's more information about this: https://cloud.google.com/appengine/docs/standard/java/datastore/entities#Ancestor_paths
I think this makes a lot of sense. I would just need to query based on the Tutor entity's key. Here's some more info about ancestor queries: https://cloud.google.com/appengine/docs/standard/java/datastore/transactions#uses_for_transactions

When we create a new TimeRange entity, we would need to specify the parent by adding the tutor entity's key as the second parameter:

`Entity time = new Entity("TimeRange", <tutor entity key>)`

I'm not sure how ancestors would work for TutorSession entities though because the sessions belong to both a tutor and a student. If I didn't use ancestors for this, I'm not sure how to use transactions for the query.

betrevisan: Oh I see! That does make a lot of sense to me. I was just worried about it still being thread-safe. I'm not entirely sure whether it would be. Do you know? If not we can discuss it during pod sync today. 

As for retrieving TutorSessions, I think we could store both the tutorID and the studentID in the session. So this would mean that when we need to get a student's upcoming tutor session, we would use their id and get all the tutor sessions that have their studentID property equal to the id we passed down. Same approach for a tutor, except that in that case we would be looking at the tutorID property.

kashiish: Yeah, I'll definitely bring it up during the pod sync. 

That's what I was thinking for TutorSessions as well, but filtering by ID is not an ancestor query so I won't be able to use transactions for it. I won't be able to use transactions for any type of filtering I think (filtering by topics, email, etc.). I'm not entirely sure if we need transactions for these types of queries but we can definitely ask about it during the meeting.

betrevisan: I'll go ahead an approve this PR since now we do not need to use transactions for this anymore. Thanks for looking into all of this :)

PR #50, opened 2020-07-08
Manage Tutor's Availability

betrevisan:
The commits proposed in this PR implement the manage-availability page, where the tutor can delete or add new available time slots to their availability, Changes include creating a servlet to handle adding new time slots and a servlet to handle deleting time slots, creating the js file with the functions for this new page, and all the necessary test files.

### DIFF
--- a/src/main/java/com/google/sps/data/SampleData.java
+++ b/src/main/java/com/google/sps/data/SampleData.java
@@ -17,6 +17,7 @@ package com.google.sps.data;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
+import java.util.Date;
 
 /** Class that stores sample Tutor objects for testing. */
 public final class SampleData {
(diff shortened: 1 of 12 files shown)


src/main/java/com/google/sps/data/SampleData.java
```diff
@@ -146,21 +147,50 @@ public static Student getStudentByEmail(String email) {
...
     }
 
+    /** Adds the given timeslot to the given tutor's availability list. */
+    public static void addAvailabilityByTutorEmail(String email, TimeRange timeslot) {
+        Tutor tutor = getTutorByEmail(email);
+        tutor.addAvailabilityByTimeRange(timeslot);
+    }
+
+    private static boolean isTimeRangeEqual(TimeRange timerange1, TimeRange timerange2) {
+        Calendar timerange1Calendar = timerange1.getDate();
```
kashiish, line 170: Instead of creating a new Calendar date, could you just compare start, end, and date of the two time ranges?

src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
```diff
@@ -0,0 +1,71 @@
+// Copyright 2019 Google LLC
```
kashiish, line 1: I noticed that you implemented the doPost method of the AvailabilityServlet for **getting** a tutor's availability. Should that implementation be in the doGet method? If you can do that, I think it would make sense to move the  AddAvailabilityServlet's doPost method to the AvailabilityServlet (kind of like how we did it for the comments in the starter project).

src/main/java/com/google/sps/servlets/DeleteServlet.java
```diff
@@ -0,0 +1,65 @@
+// Copyright 2019 Google LLC
```
kashiish, line 1: I would rename this to `DeleteAvailabilityServlet.java`

src/main/webapp/manage-availability.js
```diff
@@ -0,0 +1,106 @@
...
+function getAvailabilityManage() {
+    var queryString = new Array();
+    window.onload = readTutorID(queryString, window);
+    const tutorID = queryString["tutorID"];
+
+    document.getElementById("tutorEmail").value = tutorID;
+
+    const params = new URLSearchParams();
+    params.append('tutorID', tutorID);
+    fetch('/availability', {method: 'POST', body: params}).then(response => response.json()).then((timeslots) => {
```
kashiish, line 24: I think this would make more sense as a GET request.

betrevisan: The thing is that GET requests don't take parameters and I need to pass the tutorID as a parameter in order to load that tutor's availability.

src/main/java/com/google/sps/servlets/DeleteServlet.java
betrevisan: That sounds good thanks!

src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
betrevisan: I implemented it as a post doPost method so that I could pass down the tutor's id. I could not figure out how to pass down the tutorID in a GET request.

src/main/webapp/manage-availability.js
kashiish: You can do `/availability?<parameter name>=<value>` to pass parameters in a GET request. Then in the servlet, you can do `request.getParameter(<parameter name>)`.

src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
kashiish: I left a comment below on how to do this. Let me know if it works!

src/main/java/com/google/sps/data/SampleData.java
betrevisan: I did this and it worked well! I had to create a new PR for this because of the cloud issues I was getting but I added this in there. I'll finalize that PR soon.

src/main/webapp/manage-availability.js
betrevisan: Oh this would be great!! Thank you so mcuh

src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
betrevisan: Thanks! I'll try using it

src/main/webapp/manage-availability.js
betrevisan: I realized that the addavailability servlet is actually writing to the database so I think it should be doPost. However, my availability servlet is using a doPost function when it should be in fact just using a doGet because in that servlet it just reads the available time slots. I will create a new PR that addresses this issue if that's okay just so that this one doesn't get too complex.

PR #52, opened 2020-07-08
Integrate Datastore for scheduling a tutoring session

kashiish:
These commits create a new interface called `TutorSessionDatastoreService.java `that has methods to add a new tutoring session and get tutoring sessions by student or tutor emails. I created two new classes that implement this interface, `RealTutorSessionDatastore` which uses the actual App Engine Datastore and `MockTutorSessionDatastore` which uses the `SampleData` class. Both the `SchedulingServlet` and `ConfirmationServlet` use this service to access `TutorSessions`. 

I also added a new constructor in `SchedulingServlet` and `ConfirmationServlet` that will be used for testing. This constructor creates a new `MockTutoringSessionDatastore` instead of a `RealTutoringSessionDatastore`.


src/main/java/com/google/sps/utilities/RealTutorSessionDatastore.java
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+    * Gets all the tutor session entities for a userType (tutor or student) with the corresponding email. 
+    * @return ArrayList<TutorSession>, empty list if the student does not exist
+    */
+    private ArrayList<TutorSession> getScheduledSessions(DatastoreService datastore, String userType, String email) {
+        ArrayList<TutorSession> sessions = new ArrayList<TutorSession>();
+
+        //get all sessions with given email
+        Filter filter = new FilterPredicate(userType + "Email", FilterOperator.EQUAL, email.toLowerCase());
```
betrevisan, line 103: How does this make the distinction between the sessions a user is a tutor and the ones a user is a student?

src/main/java/com/google/sps/utilities/RealTutorSessionDatastore.java
```diff
@@ -0,0 +1,171 @@
...
+    /**
+    * Updates the TimeRange entity that the student selected to reflect that it is now scheduled and not available.
+    * @return long, the id of the TimeRange entity
+    */
+    private long updateTimeRange(String email, TimeRange time, DatastoreService datastore, Transaction txn) {
+        //filter by tutor's email and time range properties
+        CompositeFilter timeFilter = CompositeFilterOperator.and(FilterOperator.EQUAL.of("email", email),
+                                                                 FilterOperator.EQUAL.of("start", time.getStart()), 
+                                                                 FilterOperator.EQUAL.of("end", time.getEnd()), 
+                                                                 FilterOperator.EQUAL.of("date", new Gson().toJson(time.getDate())));
```
betrevisan, line 152: Is the filter operator able to compare dates using equal? When I tried comparing Calendars I had to use the compareTo method.

src/main/java/com/google/sps/utilities/RealTutorSessionDatastore.java
kashiish: The `userType` variable is going to either be "student" or "tutor". So if `userType` is "student", it will filter by "studentEmail" and if the `userType` is "tutor", it will filter by "tutorEmail".

src/main/java/com/google/sps/utilities/RealTutorSessionDatastore.java
kashiish: I'm comparing JSON strings of the date objects here, so it's just String comparison.

PR #53, opened 2020-07-08
Edit availability

betrevisan:
The commits proposed in this PR implement the manage-availability page, where the tutor can delete or add new available time slots to their availability, Changes include creating a servlet to handle adding new time slots and a servlet to handle deleting time slots, creating the js file with the functions for this new page, and all the necessary test files.

### DIFF
--- a/src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
+++ b/src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
@@ -1,0 +1,71 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import com.google.sps.data.Tutor;
+import com.google.sps.data.TimeRange;
+import com.google.sps.data.TutorSession;
+import com.google.sps.data.SampleData;
+import com.google.gson.Gson;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Calendar;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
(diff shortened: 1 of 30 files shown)


src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
```diff
@@ -0,0 +1,71 @@
...
+                                .setCalendarType("iso8601")
+                                .setDate(Integer.parseInt(year), Integer.parseInt(month), Integer.parseInt(day))
+                                .build();
+
+        TimeRange timeslot = TimeRange.fromStartToEnd(start, end, date);
+
+        // Add available timeslot
+        SampleData.addAvailabilityByTutorEmail(tutorID, timeslot);
+
+        String json = new Gson().toJson(SampleData.getSampleTutors());
```
kashiish, line 65: Why are we returning a list of tutors here?

src/main/java/com/google/sps/servlets/DeleteAvailabilityServlet.java
```diff
@@ -0,0 +1,65 @@
...
+                                .setCalendarType("iso8601")
+                                .setDate(Integer.parseInt(year), Integer.parseInt(month), Integer.parseInt(day))
+                                .build();
+
+        TimeRange timeslot = TimeRange.fromStartToEnd(Integer.parseInt(start), Integer.parseInt(end), date);
+
+        // Remove available timeslot
+        SampleData.deleteAvailabilityByTimeRange(tutorID, timeslot);
+
+        String json = new Gson().toJson(SampleData.getSampleTutors());
```
kashiish, line 60: And here?

src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
betrevisan: This is because I'm using it for testing purposes. Maybe I can just return that specific tutor's availability though. I'll try implementing that.

src/main/java/com/google/sps/servlets/DeleteAvailabilityServlet.java
betrevisan: Also for testing but I'll try to make it more specific.

src/main/webapp/manage-availability.html
```diff
@@ -0,0 +1,189 @@
...
+                    <option value="14">14</option>
+                    <option value="15">15</option>
+                    <option value="16">16</option>
+                    <option value="17">17</option>
+                    <option value="18">18</option>
+                    <option value="19">19</option>
+                    <option value="20">20</option>
+                    <option value="21">21</option>
+                    <option value="22">22</option>
+                    <option value="23">23</option>
```
samfalberg, line 72: We can definitely worry about this later but it's something to think about: Should we have hours in AM/PM? Or maybe we could do some kind of location based thing so it displays 24 hour time if you're accessing the webapp from a non-American country.

betrevisan: Yeah that would be better. For this first version I think it would be okay though. But I can definitely do this next week,

PR #56, opened 2020-07-09
Unit tests

betrevisan:
The commits proposed in this PR make the transition from having the Sample Data be static to having the sample data be non-static. Sample Data is now only a resource for testing and not  the real database. Changes were made to servlets and their test files in order to make this transition.


src/main/java/com/google/sps/servlets/HistoryServlet.java
```diff
@@ -48,7 +68,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
 
         List<TutorSession> scheduledSessions = new ArrayList<TutorSession>();
 
-        for (Student student : SampleData.getSampleStudents()) {
+        for (Student student : datastore.getStudents()) {
```
kashiish, line 71: I'll replace this with the datastore implementation after merging.

src/main/java/com/google/sps/servlets/HistoryServlet.java
```diff
@@ -19,6 +19,9 @@
 import com.google.sps.data.TutorSession;
 import com.google.sps.data.SampleData;
 import com.google.sps.data.Student;
+import com.google.sps.utilities.StudentsDatastoreService;
```
kashiish, line 22: I think you saw in my PR, I just used the TutorSession datastore service for history and rating because they basically used the same methods.

src/main/java/com/google/sps/servlets/RatingServlet.java
```diff
@@ -44,10 +64,10 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
         // Update tutor's rating
-        SampleData.rateTutorByEmail(tutorEmail, studentEmail, rating);
+        datastore.rateTutor(tutorEmail, studentEmail, rating);
 
-        String jsonTutors = new Gson().toJson(SampleData.getSampleTutors());
-        String jsonStudents = new Gson().toJson(SampleData.getSampleStudents());
+        String jsonTutors = new Gson().toJson(datastore.getTutors());
+        String jsonStudents = new Gson().toJson(datastore.getStudents());
 
         String json = new Gson().toJson(new String[]{jsonTutors, jsonStudents});
```
kashiish, line 72: I ended up just removing these lines in my PR because I don't think we use the list of tutors and students returned anywhere.

src/main/java/com/google/sps/servlets/HistoryServlet.java
betrevisan: Perfect, thanks!

src/main/java/com/google/sps/servlets/HistoryServlet.java
betrevisan: Yeah, I thought that was a better approach

src/main/java/com/google/sps/servlets/RatingServlet.java
betrevisan: They are being used inside RatingTest.java to check the if the returned contents have the new ratings that they should have. I think that if we remove these lines the tests would fail. Do you know a better way we could use to verify if the mock database was updated as it should have been? I have been just passing it down as a json to do that.

kashiish: Oh, I think the tests worked for me without these lines because we weren't using the json to test. I think we were just using to the TutorSession object to check if its properties changed.

PR #57, opened 2020-07-09
Integrating datastore into rating and history servlets

kashiish:
These commits integrate datastore for the history and rating features. I added onto the `TutorSessionDatastoreService` by creating a new `rateTutorSession` method. In order to implement this method, I added a new field to the TutorSession class which stores an id. This id is used to find a tutor session in datastore and update its rating. I updated tests and other datastore services to reflect this change. Other minor updates include disabling rating if the session is already rated and adding the `ratingSum` and `ratingCount` properties to the Tutor entity in `RegistrationServlet` (they were missing from the entity but not the class).


src/main/java/com/google/sps/data/TutorSession.java
```diff
@@ -23,16 +23,21 @@
  */
 public final class TutorSession {
 
+    //an ID counter used for creating IDs for tutor sessions
+    //used for testing only
+    private static long test_id = 0;
```
betrevisan, line 28: This is a good idea! I struggled to be able to refer to an specific tutoring session in the mock database.

src/main/java/com/google/sps/utilities/MockTutorSessionDatastore.java
```diff
@@ -65,4 +65,27 @@ public void addTutorSession(String tutorEmail, String studentEmail, TutorSession
         return student.getScheduledSessions();
     }
 
+    /**
+    * Adds the given rating to a tutor session and updates the tutor's overall rating.
```
betrevisan, line 69: I ended up creating a different interface and a different mock for rating but I think it's easier to keep the rating functions inside the TutorSessionDatastoreService. We can just use this approach after we merge.

src/main/webapp/history.js
```diff
@@ -20,6 +20,14 @@ function getTutoringSessionHistory() {
...
     params.append('studentEmail', studentEmail);
     fetch('/history', {method: 'POST', body: params}).then(response => response.json()).then((tutoringSessions) => {
+        //if there was an error
+        if(tutoringSessions.error) {
+            var container = document.getElementById('tutoringSessionHistory');
+            var errorMessage = document.createElement("p");
+            p.innerText = tutoringSessions.error;
+            container.appendChild(errorMessage);
+            return;
+        }
```
betrevisan, line 30: Nice!

src/main/java/com/google/sps/data/TutorSession.java
kashiish: Yeah, I noticed we were finding tutor sessions by tutor and student email but then I thought we could probably have more than one session with the same tutor and student emails, so an ID would work better.

betrevisan:
@googlebot I consent.

src/main/java/com/google/sps/utilities/RealTutorSessionDatastore.java
```diff
@@ -92,6 +92,66 @@ public void addTutorSession(String tutorEmail, String studentEmail, TutorSession
...
+
+            updateTutorRating(datastore, txn, tutorEmail, rating);
+
+            datastore.put(txn, sessionEntity);
+            
+        } catch (EntityNotFoundException e) {
+            success = false;
+        } finally {
+          if (txn.isActive()) {
+            txn.rollback();
```
samfalberg, line 129: What happens in the finally block here?

kashiish: This basically just ends the transactions. Even if there was an error somewhere in the try block, the finally statement makes sure the transaction ends.

PR #60, opened 2020-07-10
Remove set hour from tests

betrevisan:
The commit proposed in this PR get rid of the set hour of the day to 9 in the delete and add availability servlets.

### DIFF
--- a/src/main/java/com/google/sps/data/Student.java
+++ b/src/main/java/com/google/sps/data/Student.java
@@ -71,3 +71,4 @@ public final class Student {
 }
 
 
+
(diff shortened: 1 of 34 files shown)


src/main/java/com/google/sps/data/Student.java
```diff
@@ -10,7 +10,7 @@
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
-// limitations under the License.
+// limitations under the License.static
```
kashiish, line 13: Typo?

PR #63, opened 2020-07-13
Fix delete availability

betrevisan:
The changes proposed in this PR fix the issue of not displaying a user's availability in the manage availability page so it can be deleted. The issue was that the url component was not being encoded properly and the name of the component was tutorID when it should have been userID.


betrevisan:
> I think some of these changes were in your previous PR, but LGTM

It only shows the relevant changes now because the files that were dependent on it were merged already.

PR #64, opened 2020-07-13
Fix time slot formatting

betrevisan:
The changes proposed in this commit fix the formatting of the start hour in each time slot box. The issue was that the start hour was showing up as a float when it was not a full hour (e.g. start hour 2:30pm would be displayed as 2.5:30pm). The fix was to get the floor of each start hour.

### DIFF
--- a/src/main/java/com/google/sps/data/LoginStatus.java
+++ b/src/main/java/com/google/sps/data/LoginStatus.java
@@ -21,11 +21,13 @@ public class LoginStatus {
     private boolean needsToRegister;
     private String url;
     private String userId;
+    private String userEmail;
 
-    public LoginStatus(boolean isLoggedIn, boolean needsToRegister, String url, String userId) {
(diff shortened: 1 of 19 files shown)


src/main/java/com/google/sps/data/LoginStatus.java
```diff
@@ -21,11 +21,13 @@
     private boolean needsToRegister;
     private String url;
     private String userId;
+    private String userEmail;
```
kashiish, line 24: I think some of these changes were in your previous two PRs, can you make a new one with just timeslot formatting changes?

betrevisan: Now that the commits that this PR depended on are merged only the relevant changes show up. I can manually create a new one though.

kashiish: This works!

PR #65, opened 2020-07-13
Manage scheduled sessions

betrevisan:
The commits proposed in this PR add the page where the user can cancel any upcoming scheduled tutoring session. Changes include the creation of a servlet to handle deleting tutoring sessions and updating a tutor's availability, adding new methods to the tutor session interface and implementing them both in the real and in the mock classes, and connecting the page to manage tutoring sessions to the rest of the project through the profile page. Relevant tests are included.

### DIFF
--- a/src/main/java/com/google/sps/data/SampleData.java
+++ b/src/main/java/com/google/sps/data/SampleData.java
@@ -177,19 +177,25 @@ public final class SampleData {
     /** Adds the given TutorSession to the scheduled sessions array of the given tutor. */
     public void addToTutorScheduledSessionsByEmail(String email, TutorSession tutoringSession) {
         Tutor tutor = getTutorByEmail(email);
-
-        // tutors.remove(tutor);
         tutor.addToScheduledSessions(tutoringSession);
-        // tutors.add(tutor);
+    }
+
+    /** Deletes the given TutorSession from the scheduled sessions array of the given tutor. */
+    public void deleteFromTutorScheduledSessionsByEmail(String email, TutorSession tutoringSession) {
+        Tutor tutor = getTutorByEmail(email);
+        tutor.deleteFromScheduledSessions(tutoringSession);
     }
 
     /** Adds the given TutorSession to the scheduled sessions array of the given student. */
     public void addToStudentScheduledSessionsByEmail(String email, TutorSession tutoringSession) {
         Student student = getStudentByEmail(email);
-
-        // students.remove(student);
         student.addToScheduledSessions(tutoringSession);
-        // students.add(student);
+    }
+
+    /** Deletes the given TutorSession from the scheduled sessions array of the given student. */
+    public void deleteFromStudentScheduledSessionsByEmail(String email, TutorSession tutoringSession) {
+        Student student = getStudentByEmail(email);
+        student.deleteFromScheduledSessions(tutoringSession);
     }
 
      public void rateTutorByEmail(String tutorEmail, String studentEmail, int newRating) {
--- a/src/main/java/com/google/sps/data/SampleData.java
+++ b/src/main/java/com/google/sps/data/SampleData.java
@@ -46,7 +46,9 @@ public final class SampleData {
     private final Calendar AUGUST182020 = new Calendar.Builder()
                                                         .setCalendarType("iso8601")
                                                         .setDate(2020, 7, 18)
-                                                        .build();                                                        
+                                                        .build();
+    private final TutorSession bernardoSession = new TutorSession("btrevisan@google.com", "btrevisan@google.com", null, null, TimeRange.fromStartToEnd(540, 600, MAY182020), 0); 
+    private final TutorSession samSession = new TutorSession("sfalberg@google.com", "sfalberg@google.com", null, null, TimeRange.fromStartToEnd(540, 600, AUGUST182020), 1);                                         
 
     private ArrayList<Tutor> tutors = new ArrayList<Tutor> (Arrays.asList(
         new Tutor("Kashish Arora", "Kashish\'s bio", "images/pfp.jpg", "kashisharora@google.com", new ArrayList<String> (Arrays.asList("Math", "History")),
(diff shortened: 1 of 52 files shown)


src/main/java/com/google/sps/data/SampleData.java
```diff
@@ -177,19 +177,27 @@ public void addAvailabilityByTutorEmail(String email, TimeRange timeslot) {
     /** Adds the given TutorSession to the scheduled sessions array of the given tutor. */
     public void addToTutorScheduledSessionsByEmail(String email, TutorSession tutoringSession) {
         Tutor tutor = getTutorByEmail(email);
-
-        // tutors.remove(tutor);
         tutor.addToScheduledSessions(tutoringSession);
-        // tutors.add(tutor);
+    }
```
kashiish, line 181: Just a heads up, I don't think we'll be needing the mock datastore services anymore since I'm using the local datastore helper to test now.

src/main/java/com/google/sps/servlets/DeleteTutorSessionServlet.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    public void init() {
+        datastore = new RealTutorSessionDatastore();
+    }
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        String tutorEmail = request.getParameter("tutorEmail");
```
kashiish, line 57: I'll change this later after this merges, but since we'll be using the user id to get data now, this will have to be the ID instead of email.

src/main/java/com/google/sps/utilities/RealTutorSessionDatastore.java
```diff
@@ -232,4 +257,32 @@ private long updateTimeRange(String email, TimeRange time, DatastoreService data
         return timeEntity.getKey().getId();
     }
 
+    /**
+    * Updates the TimeRange entity that the student canceled to reflect that it is now available and not scheduled.
+    * @return long, the id of the TimeRange entity
+    */
+    private long updateTimeRangeToAvailable(String email, TimeRange time, DatastoreService datastore, Transaction txn) {
+        //filter by tutor's email and time range properties
+        CompositeFilter timeFilter = CompositeFilterOperator.and(FilterOperator.EQUAL.of("email", "scheduled"),
```
kashiish, line 266: I didn't think of this before, but this might cause some issues because there could be multiple scheduled tutoring sessions at the same time. Do you think there's another way of doing this?

I'll be replacing the email field in the TimeRange entity with a tutorID field, so maybe instead of the tutor id, we can put the tutor session id for scheduled time ranges and filter using that?

src/main/java/com/google/sps/utilities/RealTutorSessionDatastore.java
```diff
@@ -232,4 +257,32 @@ private long updateTimeRange(String email, TimeRange time, DatastoreService data
...
+
+        Query query = new Query("TimeRange").setFilter(timeFilter);
+
+        PreparedQuery pq = datastore.prepare(query);
+
+        //there should only be one result
+        Entity timeEntity = pq.asSingleEntity();
+
+        //change the email property from the tutor's email to "scheduled"
+        //instead of deleting the TimeRange entity, we can just set the email property to "scheduled" to indicate that it is a scheduled session
```
kashiish, line 279: Update comment?

src/main/java/com/google/sps/data/SampleData.java
betrevisan: Oh sounds good! I won't use them for other servlets I make then

src/main/webapp/webapptests/spec/SpecManageSessions.js
```diff
@@ -0,0 +1,107 @@
...
+    describe("when the student requests to see a their scheduled sessions", function() {
+        var mockWindow = {location: {href: "manage-sessions.html?studentEmail=test%40gmail.com", search: "?studentEmail=test%40gmail.com"}};
+
+        it("should trigger the fetch function", function() {
+            spyOn(window, "onload").and.callFake(function() {
+                readTutorID(queryString, window);
+            });
+            spyOn(window, 'fetch').and.callThrough();
+            getTutorSessionsManage(mockWindow);
+            expect(window.fetch).toHaveBeenCalledWith('/confirmation?studentEmail=undefined', {method: 'GET'});
```
kashiish, line 26: Why is the studentEmail undefined here?

src/main/java/com/google/sps/servlets/DeleteTutorSessionServlet.java
betrevisan: Oh okay, that makes sense.

src/main/webapp/webapptests/spec/SpecManageSessions.js
betrevisan: It's undefined because that is dependent on the that function that reads the studentEmail component into the queryString. That function is not being triggered here so it is not reading the url at all, that's why the studentEmail ends up being undefined. I am testing that it reads the studentEmail in a different test so the studentEmail should be correctly processed in the real version. Because this has been tested already and the goal of this test was to test the fetch function I thought it would be okay not to have to worry about the studentEmail and only about the fetch call.

src/main/java/com/google/sps/utilities/RealTutorSessionDatastore.java
betrevisan: Thanks! I just committed this.

src/main/webapp/webapptests/spec/SpecManageSessions.js
kashiish: Ok, sounds good!

src/main/java/com/google/sps/data/SampleData.java
```diff
@@ -195,6 +196,7 @@ public void addToStudentScheduledSessionsByEmail(String email, TutorSession tuto
     /** Deletes the given TutorSession from the scheduled sessions array of the given student. */
     public void deleteFromStudentScheduledSessionsByEmail(String email, TutorSession tutoringSession) {
         Student student = getStudentByEmail(email);
+        System.out.println(email);
```
samfalberg, line 199: Do we still need these print statements here? Was this to check that the email exists?

betrevisan: Yeah this is not necessary anymore. I'll remove it

PR #66, opened 2020-07-14
Distinguish tutor and student

betrevisan:
The commits proposed in this PR make sure to keep track of whether the user is a student or a tutor. Changes include adding two extra fields to the LoginStatus object class that will be true if the user is a tutor or a student. These fields are loaded when the LoginStatusServlet is called and the user is logged in and registered. Being able to tell whether the user is a student or a tutor allows the platform to display the correct buttons for the current user's role.


src/main/java/com/google/sps/data/LoginStatus.java
```diff
@@ -22,12 +22,16 @@
     private String url;
     private String userId;
     private String userEmail;
+    private boolean tutor;
+    private boolean student;
 
-    public LoginStatus(boolean isLoggedIn, boolean needsToRegister, String url, String userId, String userEmail) {
+    public LoginStatus(boolean isLoggedIn, boolean needsToRegister, String url, String userId, String userEmail, boolean tutor, boolean student) {
```
kashiish, line 28: I think we really only need one of these booleans here. For example, if we kept just the `tutor` boolean, we could check if the user is a tutor by checking if `tutor` is true or we could check if the user is a student by checking if `tutor` is false.

src/main/java/com/google/sps/servlets/LoginStatusServlet.java
```diff
@@ -96,4 +107,24 @@ public String getName(String userId, DatastoreService datastore) {
...
     }
+
+    public boolean isTutor(String userId, DatastoreService datastore) {
+        Filter filter = new FilterPredicate("userId", FilterOperator.EQUAL, userId);
+        Query query = new Query("Tutor").setFilter(filter);
+        PreparedQuery results = datastore.prepare(query);
+
+        Entity tutorEntity = results.asSingleEntity();
+
+        return tutorEntity != null;
```
kashiish, line 118: I think there is a "role" property in the User entity that will tell you if the user is a tutor or a student.

src/main/java/com/google/sps/data/LoginStatus.java
betrevisan: The thing is that when we start having the possibility of a user being both a tutor and a student, we would have to have the two different fields. Otherwise, someone would have to create two different accounts, one for tutoring and one for learning, which I do not think is the best approach. Right now, I was envisioning having a tab (similar to the one that makes the distinction between book and tutors in the search area) to display the information of a user's learning activities and a tab for their tutoring activities. Of course, if the user is only a tutor or a student then that tab would not be displayed.

src/main/java/com/google/sps/servlets/LoginStatusServlet.java
betrevisan: Oh cool! I'll look into it

kashiish: Unless we're accounting for people that are both a student and a tutor? In that case, should we make the "role" property equal to something like "both"?

src/main/java/com/google/sps/data/LoginStatus.java
kashiish: Oh yeah, that makes sense!

src/main/webapp/registration.js
```diff
@@ -57,23 +57,32 @@ function displayLoginLogoutLinkHelper(document, loginStatus) {
...
-        });
-        document.getElementById('tutor-session-settings').addEventListener('click', () => {
-            redirectToManageSessions(window, loginStatus);
-        });
-        document.getElementById('history').addEventListener('click', () => {
-            redirectToHistory(window, loginStatus);
-        });
+
+        // If the user is a tutor, display availability settings
+        if (loginStatus.tutor) {
```
kashiish, line 67: What happens here if the user is both a tutor and a student?

src/main/java/com/google/sps/data/LoginStatus.java
betrevisan: I just go rid of this now because I'm using role.

src/main/java/com/google/sps/servlets/LoginStatusServlet.java
betrevisan: Yep,once I implement the possibility of registering as both student and tutor I'll make sure to add the distinction here and add the tab to go from the tutor's view to the student's view.

src/main/webapp/registration.js
betrevisan: After I implement the possibility of registering as both student and tutor, there will be a tab to see the student's view and the tutor's view. But since allowing the student to register as both student and tutor would involve making changes to the registration servlet and others, I'm doing that in a different PR just so that it is more organized.

kashiish: Sounds good!

src/main/java/com/google/sps/servlets/LoginStatusServlet.java
```diff
@@ -96,4 +106,18 @@ public String getName(String userId, DatastoreService datastore) {
...
+        PreparedQuery results = datastore.prepare(query);
+
+        Entity tutorEntity = results.asSingleEntity();
+
+        if (tutorEntity == null) {
+            return null;
+        } else{
+            return (String) tutorEntity.getProperty("role");
+        }
+    }
```
kashiish, line 122:
```suggestion
        Entity userEntity = results.asSingleEntity();

        if (userEntity == null) {
            return null;
        } else{
            return (String) userEntity.getProperty("role");
        }
    }
```

betrevisan: thanks! I added it as a new commit.

PR #67, opened 2020-07-14
Verify user permissions

betrevisan:
The commits proposed in this PR add permission checks to the backend servlets. Changes include making sure the user who is trying to delete or add availability is the actual tutor who will be affected by it. Additionally, it also checks whether the user trying to rate, schedule, or cancel a tutoring session is the student who is actually connected to this tutoring session.

### DIFF
--- a/src/main/java/com/google/sps/utilities/RealAvailabilityDatastore.java
+++ b/src/main/java/com/google/sps/utilities/RealAvailabilityDatastore.java
@@ -75,7 +75,9 @@ public final class RealAvailabilityDatastore implements AvailabilityDatastoreSer
         UserService userService = UserServiceFactory.getUserService();
         // If the email of the currently logged in user does not match the email of the user whose
         // availability will be altered, stop the request.
-        if (userService.getCurrentUser().getEmail().toLowerCase() != email.toLowerCase()) {
+        System.out.println(userService.getCurrentUser().getEmail().toLowerCase());
+        System.out.println(email.toLowerCase());
+        if (!userService.getCurrentUser().getEmail().toLowerCase().equals(email.toLowerCase())) {
             System.out.println("The current user does not have permission to add availability");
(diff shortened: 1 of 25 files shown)


src/main/java/com/google/sps/utilities/RealAvailabilityDatastore.java
```diff
@@ -32,6 +32,8 @@
 import com.google.appengine.api.datastore.Query.CompositeFilter;
```
kashiish, line 32: Should this stuff be handled by session tokens? I was going to start working on that soon.

betrevisan: I thought this would still be a helpful check event after that is implemented just so that we can avoid injections. But maybe it can be deleted after sessions are implemented if they are able to handle this.

src/main/java/com/google/sps/utilities/RealAvailabilityDatastore.java
```diff
@@ -70,6 +72,16 @@
     */
     @Override
     public boolean addAvailability(String email, TimeRange time) {
+        UserService userService = UserServiceFactory.getUserService();
+        // If the email of the currently logged in user does not match the email of the user whose
+        // availability will be altered, stop the request.
+        System.out.println(userService.getCurrentUser().getEmail().toLowerCase());
+        System.out.println(email.toLowerCase());
```
kashiish, line 79: Can we delete these?

betrevisan: Just delete them

betrevisan:
> LGTM. Since I replaced all the email properties with ID, we can switch that in these files after my PR is merged.

Sounds good! I'll wait for your branch to be merged before I do then.

PR #68, opened 2020-07-14
Use local datastore helper for testing and use user IDs to query into database

kashiish:
Sorry this is such a big PR! I had to change little things in a lot of different files in order to switch to ID querying. 

These commits reflect major changes including switching all servlet testing to use the LocalServiceTestHelper datastore and all datastore query filters to use user IDs instead of emails. I got rid of the interfaces completely, so now the "real" datastore implementations are in the "DatastoreService" classes. The majority of the changes in the datastore service classes are old methods from the "RealDatastore" classes. I also updated the SampleData class to add a new method that will transfer all of the Tutor objects into datastore so we can use them for testing. Lastly, I updated all the JS fetch calls and Jasmine tests to reflect the changes with the id/email. 

Let me know if you have any questions because there are a lot of changes!


src/main/java/com/google/sps/data/SampleData.java
```diff
@@ -47,44 +52,45 @@
                                                         .setCalendarType("iso8601")
                                                         .setDate(2020, 7, 18)
                                                         .build();
-    private final TutorSession bernardoSession = new TutorSession("btrevisan@google.com", "btrevisan@google.com", null, null, TimeRange.fromStartToEnd(540, 600, MAY182020), 0); 
-    private final TutorSession samSession = new TutorSession("sfalberg@google.com", "sfalberg@google.com", null, null, TimeRange.fromStartToEnd(540, 600, AUGUST182020), 1);                                         
+    //9 and 14 are the ids local datastore gives to these entities
+    private final TutorSession bernardoSession = new TutorSession("1", "1", null, null, TimeRange.fromStartToEnd(540, 600, MAY182020), 9); 
+    private final TutorSession samSession = new TutorSession("2", "2", null, null, TimeRange.fromStartToEnd(540, 600, AUGUST182020), 14);
```
betrevisan, line 57: How does it decide on 9 and 14?

src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
```diff
@@ -19,37 +19,24 @@
...
 import java.util.List;
 import java.util.Calendar;
+import java.util.Optional;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-@WebServlet("/addAvailability")
+@WebServlet("/add-availability")
```
betrevisan, line 34: Is hyphenation the default way to name servlets?

src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
```diff
@@ -60,7 +47,8 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
 
     @Override
     public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-        String tutorID = request.getParameter("tutorEmail");
+        //Set default value to -1 
+        String tutorID = Optional.ofNullable(request.getParameter("tutorID")).orElse("-1");
```
betrevisan, line 51: Is there ever a case in which the id would end up being -1? Isn't it handles by the datastore itself?

src/main/java/com/google/sps/data/SampleData.java
kashiish: I'm 90% sure the id is assigned by what order the entity is put into datastore. For these TutorSession entities, the first one is the 9th entity that gets put into datastore and the second one is the 14th. This is only relevant to the local datastore and SampleData. So if we decide to change SampleData and add more entities before these two session, we would need to update the ids.

src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
kashiish: The DeleteAvailabilityServlet used hyphenation, so I changed this one too. I'm not sure if there is a default way.

src/main/webapp/manage-availability.js
```diff
@@ -12,13 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/** A function that adds event listeners to a DOM objects. */
+function addEventListeners() {
+    document.getElementById("timeslot-form").addEventListener('submit', event => {
+        event.preventDefault();
```
betrevisan, line 18: What does this preventDefault do?

src/main/webapp/manage-availability.js
```diff
@@ -91,6 +97,28 @@ function createTimeSlotBoxManage(timeslot, tutorID) {
...
+    params.append('month', document.getElementById('month').value);
+    params.append('year', document.getElementById('year').value);
+    params.append('tutorID', userID);
+
+    fetch('/add-availability', {method: 'POST', body: params}).then(() => {
+        window.location.href = "/manage-availability.html?userID=" + userID;
+    });
+
+}
+
```
betrevisan, line 121: I was a using a post request on the form to add the availability but this looks like a good way to do it too. Was it mostly for readability or was there another reason for choosing this approach?

src/main/webapp/webapptests/spec/SpecConfirmation.js
```diff
@@ -60,7 +61,13 @@ describe("Confirmation", function() {
         });
 
         it("should have the inner HTML of the h3 tag equal to to the email of the tutor", function() {
-            expect(actual.childNodes[0].childNodes[0].innerHTML).toEqual("Tutoring Session with tester@gmail.com");
+            var tutor = {email: "tester@gmail.com"};
+            spyOn(window, "fetch").and.returnValues(Promise.resolve({json: () => Promise.resolve(user)}), Promise.resolve({json: () => Promise.resolve(tutor)}));
```
betrevisan, line 65: How does this Promise.resolve() work? I hadn't seen it before

src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
kashiish: I think Datastore only assigns positive IDs. They advise against using negative IDs in datastore (https://cloud.google.com/datastore/docs/best-practices#keys), so I don't think Datastore would ever generate a negative ID. In this case, if the tutorID is null, it will be assigned -1 for the ID. I'll add a check to make sure that if the ID is -1, it should not create a new availability.

src/main/webapp/manage-availability.js
kashiish: It prevents the form from submitting and reloading the page.

src/main/webapp/manage-availability.js
kashiish: I couldn't send the tutorID when we were using the post request on the form, but I agree that way was definitely easier.

src/test/java/com/google/sps/AvailabilityTest.java
```diff
@@ -49,20 +52,35 @@
...
-        servlet = new AvailabilityServlet(true);
-        TutorSession.resetIds();
+    public void setUp() {
+        helper.setUp();	        
+
+        servlet = new AvailabilityServlet();
+        servlet.init();
+
+        SampleData sample  = new SampleData();
+        sample.addTutorsToDatastore();
```
betrevisan, line 68: When this is added the changes from previous tests are not carried over, are they? Is the sample data being added the its initial version?

src/test/java/com/google/sps/DeleteTutorSessionTest.java
```diff
@@ -86,21 +114,30 @@ public void testDoPost() throws Exception {
...
+    //     Entity timeEntity = new Entity("TimeRange");
+
+    //     //the id of the session we are testing
+    //     timeEntity.setProperty("tutorID", "14");
+    //     timeEntity.setProperty("start", time.getStart());
+    //     timeEntity.setProperty("end", time.getEnd());
+    //     timeEntity.setProperty("date", new Gson().toJson(time.getDate()));
+
+    //     datastore.put(timeEntity);
+    // }
```
betrevisan, line 141: Can this be deleted?

src/main/java/com/google/sps/data/SampleData.java
```diff
@@ -14,9 +14,14 @@
...
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
+import com.google.gson.Gson;
+
 
 /** Class that stores sample Tutor objects for testing. */
 public final class SampleData {
```
betrevisan, line 27: Could all of the of the previous methods in sample data like getstudentbyemail or deletestudentbyemail be deleted now that they are not being used anymore?

src/main/java/com/google/sps/data/SampleData.java
betrevisan: Oh okay, that makes sense

betrevisan: The would probably account for this right? I mean if we add more entities and the id number changes

src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
betrevisan: yep, that sounds good. I'll start using hyphenation too:)

src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
betrevisan: Cool! It might be a good idea to add that check to the other places where the id is being set to -1 too

src/main/webapp/manage-availability.js
betrevisan: ohhh this is helpful

src/main/java/com/google/sps/data/SampleData.java
kashiish: Yeah, I'm not really sure of the best way to handle this besides manually changing the ids every time we add something to SampleData. It shouldn't affect the ids unless we add new entities **before** these tutoring sessions.

src/main/java/com/google/sps/servlets/AddAvailabilityServlet.java
kashiish: Yep, just made a commit for these checks.

src/main/webapp/webapptests/spec/SpecConfirmation.js
kashiish: It basically returns a new resolved promise object. You can get the value inside the promise using .then(). This might explain it a bit better: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve

src/test/java/com/google/sps/AvailabilityTest.java
kashiish: They are not carried over. Since the local datastore helper keeps all the data in memory, the memory is wiped after the end of each test: https://cloud.google.com/appengine/docs/standard/java/tools/localunittesting#datastore-memcache

Yes, the sample data being added is the initial version (unaltered).

src/main/java/com/google/sps/data/SampleData.java
kashiish: Yes, I think so. I'll make a new commit deleting all the unused methods.

kashiish: I think we still need the getStudentByEmail and getTutorByEmail methods because a few tests still use them, but I'll delete everything else.

src/main/webapp/webapptests/spec/SpecConfirmation.js
betrevisan: thanks!

src/test/java/com/google/sps/AvailabilityTest.java
betrevisan: Perfect!

src/main/java/com/google/sps/data/SampleData.java
betrevisan: Cool, sounds good!

PR #69, opened 2020-07-14
My students page

betrevisan:
The commits proposed in this PR add a page where the tutor can see their current students. Changes include adding a new page called -mystudents.html that loads the tutor's students through a new servlet called GetStudentsServlet.java. I also added tutors as a property to the student entity. This property is responsible for keeping track of the id of the tutors who have taught a certain user. For testing purposes, I added method that add sample data students to the datastore. 

The next step will be to load the progress.html page, which will display the the progress of a student and their previous experiences. This will be in my next PR.


src/main/java/com/google/sps/servlets/GetStudentsServlet.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.List;
+import java.util.ArrayList;
+import java.lang.String;
+import com.google.gson.Gson;
+import com.google.sps.utilities.GetStudentsService;
+
+/** Servlet that returns a list of students for a given tutor. */
+@WebServlet("/my-students")
+public class GetStudentsServlet extends HttpServlet {
+    private GetStudentsService datastore;
```
kashiish, line 34: Maybe we can name this something like StudentDatastoreService so that if we need to add more student related methods, we can add them all here.

src/main/java/com/google/sps/servlets/ProfileServlet.java
```diff
@@ -70,16 +71,21 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         if (role.toLowerCase().equals("student")) {
             query = new Query("Student").setFilter(new Query.FilterPredicate("userId", Query.FilterOperator.EQUAL, userId));
             results = datastore.prepare(query);
+            System.out.println(results);
```
kashiish, line 74: Delete?

src/main/java/com/google/sps/servlets/RegistrationServlet.java
```diff
@@ -91,6 +91,8 @@ public void createStudentEntityAndPutInDatastore(DatastoreService ds, Entity ent
     entity.setProperty("pfp", pfp);
     entity.setProperty("email", email);
     entity.setProperty("learning", topics);
+    List<String> tutors = new ArrayList<String>();
+    entity.setProperty("tutors", tutors);
```
kashiish, line 95: I think adding an empty list to datastore will always add it as a null value. We can probably delete these lines and just set the property like you did in TutorSessionDatastoreService.

src/main/java/com/google/sps/utilities/GetStudentsService.java
```diff
@@ -0,0 +1,75 @@
...
+/** Accesses datastore to get students that have had a tutoring session with the given tutor. */ 
+public final class GetStudentsService {
+
+    /**
+    * Gets all students that the given userId was or will be a tutor for.
+    * @return ArrayList<Student>, empty list if no students
+    */
+    public ArrayList<Student> getStudentsForTutor(String tutorId) {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        Query studentQuery = new Query("Student");
```
kashiish, line 49: Don't know if this is necessarily easier but I'm thinking about if we have a lot of student users, this might get really slow because we would have to look through all of the students to find matching tutor ids. We could try to query TutorSessions by the tutorId and get the corresponding studentId for each session to get all the students.

src/main/webapp/my-students.js
```diff
@@ -0,0 +1,70 @@
...
+        }
+        students.forEach((student) => {
+            document.getElementById('students').appendChild(createStudentBox(student));
+        })
+    });
+}
+// Referenced to https://www.aspsnippets.com/Articles/Redirect-to-another-Page-on-Button-Click-using-JavaScript.aspx#:~:text=Redirecting%
+// 20on%20Button%20Click%20using%20JavaScript&text=Inside%20the%20Send%20JavaScript%20function,is%20redirected%20to%20the%20URL on June 23rd.
+// This function reads the id of the tutor that the student has selected, which is passed as an URI component, and add it to the queryString array..
+function readTutorID(queryString, window) {
```
kashiish, line 35: We don't have to do this now but I think a general JS file would be helpful for methods like this because we use it in a lot of files (along with getUsers, getUserId, displayLoginLogoutLink, etc).

src/main/java/com/google/sps/servlets/GetStudentsServlet.java
betrevisan: Makes sense. I'll do that

src/main/java/com/google/sps/servlets/RegistrationServlet.java
betrevisan: That's true! It works

src/main/java/com/google/sps/utilities/GetStudentsService.java
betrevisan: Yes, my first instinct was to look at the TutorSessions only and get the corresponding studentIds. The issue I found then was that  I'd have to look at all students to see which students have ids that match the ids  in the list of studentIds and then get their information such as (name, email,..). So this turned out to be even slower because I would not only look at all students but I'd also look at all TutorSessions.

src/main/webapp/my-students.js
betrevisan: That's I good idea. I put it as a task for me to create that file so I'll do that in a separate PR.

PR #70, opened 2020-07-15
Customize Profiles

samfalberg:
These commits will allow the user to update the info on their own profiles. This will be done by updating the values of the user's entity and storing it in the datastore again. If a user is looking at somebody else's profile, they will not have the option to update that information. 
First commit is "repeated" because I thought it didn't go through, sorry.


src/main/java/com/google/sps/servlets/ProfileServlet.java
```diff
@@ -106,7 +111,65 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    }
+
+    /**
+    * Updates a student entity and puts it in datastore, used for testing
+    */
+    public void updateStudentEntityAndPutInDatastore(DatastoreService ds, Entity entity, String userId, String bio, List<String> topics) {
+        Query query = new Query("Student").setFilter(new Query.FilterPredicate("userId", Query.FilterOperator.EQUAL, userId));
+        PreparedQuery results = ds.prepare(query);
+        entity = results.asSingleEntity();
+        entity.setProperty("bio", bio);
```
kashiish, line 159: Should we add a try/catch here in case the query returns null?

src/main/java/com/google/sps/servlets/ProfileServlet.java
```diff
@@ -106,7 +111,65 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    }
 
+    /**
+    * Updates a tutor entity and puts it in datastore, used for testing
+    */
+    public void updateTutorEntityAndPutInDatastore(DatastoreService ds, Entity entity, String userId, String bio, List<String> topics) {
+        Query query = new Query("Tutor").setFilter(new Query.FilterPredicate("userId", Query.FilterOperator.EQUAL, userId));
+        PreparedQuery results = ds.prepare(query);
+        entity = results.asSingleEntity();
+        entity.setProperty("bio", bio);
```
kashiish, line 171: Here too

src/main/webapp/profile.js
```diff
@@ -60,14 +60,28 @@ function createProfileDiv(user) {
     profileEmail.innerHTML = user.email;
 
     const profileTopics = document.createElement('p');
+    var isStudent;
     // Students and tutors have different properties, check to see what kind of profile we're displaying
     if (user.learning != undefined) {   // User is a student
```
kashiish, line 65: I think the login status now returns a "role" property which specifies if the user is a "student" or a "tutor". Not sure if that would be easier to use here.

src/test/java/com/google/sps/ProfileTest.java
```diff
@@ -120,4 +137,38 @@ public void doGetReturnsCorrectResponseForTutor() throws Exception {
...
         Assert.assertTrue(stringWriter.toString().contains(expected));
     }
+
+    @Test
+    public void doPostCancelsFormSubmit() throws Exception {
+        when(request.getParameter("submit")).thenReturn("Cancel");
+
+        servlet.doPost(request, response);
+
+        ArgumentCaptor<String> url = ArgumentCaptor.forClass(String.class);
```
kashiish, line 147: What is an ArgumentCaptor used for?

src/main/java/com/google/sps/servlets/ProfileServlet.java
```diff
@@ -106,7 +111,65 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     @Override
-    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {  
+        String userId = userService.getCurrentUser().getUserId();
+        
+        // Don't update profile if cancel button is clicked
+        if (request.getParameter("submit").equals("Cancel")) {
+            response.sendRedirect("/profile.html?userID=" + userId);
+            return;
+        }
```
betrevisan, line 121: Is it necessary to submit the post request when clicking cancel? Wouldn't the form just not be submitted?

src/main/java/com/google/sps/servlets/ProfileServlet.java
samfalberg: For some reason the compiler isn't accepting an EntityNotFoundException, so for now I'm just returning early if the query returns null.

src/main/webapp/profile.js
samfalberg: That's much easier, I add that in a new commit

src/test/java/com/google/sps/ProfileTest.java
samfalberg: It captures argument values, in this case for the sendRedirect, so I can check that the servlet is redirecting where I want it to.

src/main/java/com/google/sps/servlets/ProfileServlet.java
samfalberg: Since the cancel button is part of the form I think it submits no matter what. I tried making it a regular button that just hides the form when clicked, but that submitted the form anyway. This is the easiest thing I could find from the Google searches I did.

PR #73, opened 2020-07-16
Use sessions and servlet filters to prevent users from viewing other user's pages

kashiish:
These commits use sessions to make sure the user can only access their own pages (availability, sessions, history). By using sessions, we can prevent users from typing in a random user id as a query parameter and viewing someone else's personal data. I created a servlet filter that will intercept a request call and check if the session is valid. If it is not, it will redirect to the homepage. I also added a security constraint which will redirect users that are not logged in to log in when they try to visit personal pages. 

Lastly, I fixed up some tests and added a few tests to make sure the pages redirect when appropriate.

### DIFF
--- a/src/main/java/com/google/sps/filters/SessionFilter.java
+++ b/src/main/java/com/google/sps/filters/SessionFilter.java
@@ -32,12 +32,16 @@ import java.util.Optional;
 import com.google.appengine.api.users.UserService;
 import com.google.appengine.api.users.UserServiceFactory;
 
-/** Servlet filter that checks if the user has valid session to make a request. */
+/** Servlet filter that checks if the user has valid session to make a request.*/
 public class SessionFilter implements Filter {
 
     private ServletContext context;
     private FilterConfig filterConfig;
 
+    /**
+    * This method is called automatically before the request is sent to the servlet (specified in the web.xml file) to make sure
+    * the session is valid.
+    */
     public void doFilter(ServletRequest request, ServletResponse response, FilterChain filterChain) throws IOException, ServletException {
 
         HttpServletRequest req = (HttpServletRequest) request;
--- a/src/main/java/com/google/sps/filters/SessionFilter.java
+++ b/src/main/java/com/google/sps/filters/SessionFilter.java
@@ -62,7 +62,6 @@ public class SessionFilter implements Filter {
 
         //user has not logged in, no cookies have been created
         if(cookies == null) {
-            System.out.println("wrong 1");
             this.context.log("Unauthorized access request");
             return false;
         }
@@ -80,7 +79,6 @@ public class SessionFilter implements Filter {
 
         //The sessionId from the cookie has ".node0" as a suffix
         if(session == null || !(session.getId()+".node0").equals(sessionId)) {
-            System.out.println("wrong 2");
             this.context.log("Unauthorized access request");
 			
             return false;
--- a/src/main/java/com/google/sps/filters/SessionFilter.java
+++ b/src/main/java/com/google/sps/filters/SessionFilter.java
@@ -1,0 +1,103 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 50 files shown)


src/main/java/com/google/sps/filters/SessionFilter.java
```diff
@@ -0,0 +1,101 @@
...
+        //user has not logged in, no cookies have been created
+        if(cookies == null) {
+            this.context.log("Unauthorized access request");
+            return false;
+        }
+
+        String sessionId = "";
+        
+        for(Cookie cookie : cookies) {
+            if(cookie.getName().equals("JSESSIONID")) {
```
betrevisan, line 72: Is it really JSEESSIONID or should we name is SESSIONID?

src/main/java/com/google/sps/servlets/LogoutServlet.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+
+/** Servlet that invalidates the user's session after they log out. */
+@WebServlet("/logout")
+public class LogoutServlet extends HttpServlet {
+
+    @Override
+    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        if(request.getSession(false) != null) {
```
betrevisan, line 31: What is the false in getSession(false) for?

src/main/java/com/google/sps/filters/SessionFilter.java
kashiish: JSESSIONID is the default name the session gives to this cookie.

src/main/java/com/google/sps/servlets/LogoutServlet.java
kashiish: It lets us get the current active session (if there is one) instead of creating a new one. getSession()/getSession(true) will always create a new one.

src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,69 @@
+<web-app xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee
```
betrevisan, line 3: What are the changes in this file for?

src/test/java/com/google/utilities/TestUtilities.java
```diff
@@ -0,0 +1,37 @@
...
+    * Sets the mock request's session id.
+    */
+    public static void setSessionId(HttpServletRequest request, String id) {
+        HttpSession session = mock(HttpSession.class);
+
+        when(request.getSession(false)).thenReturn(session); 
+        when(session.getAttribute("userId")).thenReturn(id);  
+    }
+
+}
```
betrevisan, line 37: Cool!  Note to add other recurrent test functions to this later.

src/main/java/com/google/sps/filters/SessionFilter.java
```diff
@@ -0,0 +1,101 @@
...
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import javax.servlet.http.Cookie;
+import javax.servlet.ServletContext;
+import java.util.stream.*;
+import java.util.Optional;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+
+/** Servlet filter that checks if the user has valid session to make a request. */
```
betrevisan, line 35: Maybe add a comment about this being called by the files in the web.xml automatically.

src/main/webapp/WEB-INF/web.xml
kashiish: This file adds the security constraint so that it will redirect the user that is not logged in to log in if they try to access personal pages. It also specifies the servlets/url-patterns which the SessionFilter should be applied to.

PR #78, opened 2020-07-20
Page access fixes

kashiish:
These commits change a variety of things. First, I created a separate servlet for adding goals and experiences so that I could apply the session filter to those servlets. Previously, adding a goal and experience were a POST request in the Goal and Experience servlets. Since I can only apply the session filter to servlets and not requests, I needed to separate the add goal and experience POST requests to their own servlets. Next, I created a new ManageAvailability servlet that is used for the manage-availability page. Previously, the manage availability page made requests to the Availability servlet, but the availability page from the search results also used this servlet so I could not apply the session filter. Now, the add availability and get availability requests for the manage availability page are passed to the session filter to make sure the user is logged in and the correct user is trying to access the information. Lastly, I removed the security constraints from the web.xml because when someone is redirected to log in and they have not registered before, they are not redirected to the registration page after they sign in with Google. However, they are still redirected back to the page they were on before. This was a bug because then the User did not exist in our database. Now, when a user is not logged in and they try to schedule a tutoring session or access personal pages, an alert is shown telling the user that they must be signed in to view/schedule/edit and the user is redirected to the homepage.


src/main/java/com/google/sps/servlets/AddExperienceServlet.java
```diff
@@ -0,0 +1,66 @@
...
+        String studentID = Optional.ofNullable((String)request.getSession(false).getAttribute("userId")).orElse("-1");
+        String experience = request.getParameter("experience");
+
+        if(studentID.equals("-1")) {
+            response.setContentType("application/json");
+            response.getWriter().println("{\"error\": \"There was an error adding experience.\"}");
+            return;
+        }
+
+        Experience newExperience = new Experience(studentID, experience);
```
samfalberg, line 55: Can you fill me in on experiences? I think I missed when this was added.

src/main/webapp/progress.js
```diff
@@ -336,7 +342,14 @@ function deleteGoal(goal, window) {
...
 
-    fetch('/delete-goal', {method: 'POST', body: params});
+    fetch('/delete-goal', {method: 'POST', body: params}).then((response) => {
+        //if the student id is not the id of the current user
+        if(response.redirected) {
+            window.location.href = response.url;
+            alert("You must be signed in to delete a goal.");
+            return;
+        }
+    });;
```
samfalberg, line 352: Remove extra semicolon?

src/main/webapp/progress.js
```diff
@@ -358,6 +377,13 @@ function deleteExperience(experience, window) {
...
 
-    fetch('/delete-experience', {method: 'POST', body: params});
+    fetch('/delete-experience', {method: 'POST', body: params}).then((response) => {
+        //if the student id is not the id of the current user
+        if(response.redirected) {
+            window.location.href = response.url;
+            alert("You must be signed in to delete an experience.");
+            return;
+        }
+    });;
```
samfalberg, line 387: Here too

src/main/java/com/google/sps/servlets/AddExperienceServlet.java
betrevisan: I added experiences and goals last week to the progress tracker feature. Experiences keep track of any previous experiences a user might have had on the topics they are studying (past classes, courses, certificates,..) and goals keep track of goals the student sets for themselves. These are displayed to the tutor so that they can prepare a tutoring sessions that suits the student.

src/main/webapp/confirmation.js
```diff
@@ -22,7 +22,8 @@ async function getScheduledSessionsHelper(window) {
     await fetch('/confirmation', {method: 'GET'}).then((response) => {
         //if the student id is not the id of the current user
         if(response.redirected) {
-            window.location.href = response.url
+            window.location.href = response.url;
+            alert("You must be signed in to view upcoming session.");
```
betrevisan, line 26: Nice

PR #80, opened 2020-07-21
Expand topics options

samfalberg:
These commits will give the user more topics to choose from, as well as more freedom to choose custom topics. The user will have 10 default topics to choose from, and can also add any other topics in a comma separated list.


src/main/webapp/profile.html
```diff
@@ -44,10 +44,26 @@
...
                 <label for="english"> English</label><br>
-                <input type="checkbox" id="other-subject" name="other" value="Other">
-                <label for="other-subject"> Other</label><br><br>
+                <input type="checkbox" id="spanish" name="spanish" value="Spanish">
+                <label for="spanish"> Spanish</label><br>
+                <input type="checkbox" id="french" name="french" value="French">
+                <label for="french"> French</label><br>
+                <input type="checkbox" id="chinese" name="chinese" value="Chinese">
+                <label for="chinese"> Chinese</label><br>
+                <label for="other-subject">Please add any other topics in a comma separated list (e.g. "History, Business, Computer Science"):</label><br>
```
betrevisan, line 65: It's good that you added an example of how the comma separated list should look like

src/main/webapp/profile.js
```diff
@@ -59,28 +59,13 @@ function createProfileDiv(user) {
...
     fetch('/login-status').then(response => response.json()).then((loginStatus) => {
-        role = loginStatus.role;
-
-        if (loginStatus.userId == getIdParameter(window)) {
-            document.getElementById('edit-profile-btn').style.display = 'block';
-            document.getElementById('edit-profile-btn').addEventListener('click', () => {
-                editProfile(user, role, document); 
-            });
-        }
+        profileTopics.innerHTML = fetchStatusHelper(user, loginStatus, window, document);
```
betrevisan, line 66: This looks much cleaner :)

PR #85, opened 2020-07-22
Merge progress into profile

betrevisan:
The commits proposed in this PR add the progress tracker features to the profile page and delete the page that was previously designed for the progress tracker. Changes include loading the progress tracker from the profile.js file, adding permission checks on who can see/edit certain sensitive content inside the profile page, and updating jasmine tests.


src/main/webapp/progress.js
```diff
@@ -15,18 +15,15 @@
...
-function loadProgressHelper(document, loginStatus) {
-    getExperiences(document, loginStatus);
-    getGoals(document, loginStatus);
+    document.getElementById('progress-tracker').style.display = 'block';
+
+    getExperiences(document, loginStatus, user);
+    getGoals(document, loginStatus, user);
     getAchievements(document, loginStatus);
-    getPastSessionsAndTopics(document, loginStatus);
+    getPastSessionsAndTopics(document, loginStatus, user);
```
kashiish, line 26: I think I missed this but when/where do we get the `user` from?

betrevisan: The user is retrieved when the displayProfile function is called and then from there it is passed down to the function that need it.

PR #86, opened 2020-07-22
Fix Google sign out issue

kashiish:
These commits fix the issue when a user logs out of the application, Google signs out of everything. Instead of clicking a logout URL, the logout servlet deletes the authentication cookie associated with their log in. This logs them out of just Sullivan and not anything else.


src/main/java/com/google/sps/servlets/LogoutServlet.java
```diff
@@ -21,16 +21,55 @@
...
 
 /** Servlet that invalidates the user's session after they log out. */
 @WebServlet("/logout")
 public class LogoutServlet extends HttpServlet {
 
     @Override
-    public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        //log the user out if we're on the development server
+        if (SystemProperty.environment.value() != SystemProperty.Environment.Value.Production) {
```
betrevisan, line 36: Does the development server here refer to the deployed or local version? Also, what does .Production do?

src/main/java/com/google/sps/servlets/LogoutServlet.java
```diff
@@ -21,16 +21,55 @@
...
+            return;
+        }
+        
+        for(Cookie cookie : cookies) {
+            //Instead of the user clicking the logout link generated by the Authentication API
+            // we can just delete the cookies associated with the log in so that the user isn't logged out of
+            // Google completely.
+            // Reference: http://ptspts.blogspot.com/2011/12/how-to-log-out-from-appengine-app-only.html
+            if(cookie.getName().equals("dev_appserver_login") || cookie.getName().equals("SACSID")) {
+                cookie.setValue("");
```
betrevisan, line 61: I might be wrong, but today during the podsync when you said that js interprets an empty string as the homepage, does that apply to this cookie being set to an empty string?

src/main/java/com/google/sps/servlets/LogoutServlet.java
kashiish: SystemProperty.Environment.Value.Development is the local version and SystemProperty.Environment.Value.Production is the deployed version.

src/main/java/com/google/sps/servlets/LogoutServlet.java
kashiish: I don't think so. I think that only applied to links.

PR #87, opened 2020-07-22
Calendar

samfalberg:
These commits will use the Google charts API to make a calendar that will display on the manage-sessions and manage-availability page. The frontend will fetch data from the backend and use that to create sessions to display on the calendar. On manage-sessions, the student will be able to see which tutor they're meeting with, what subject they're learning, and the time range of the session. On manage-availability, the tutor will be able to see which days they have available sessions, and what time of day these sessions take place.


samfalberg:
Here's what the calendar currently looks like (with hardcoded data). Ideally, we will format it so it nicely displays an entire weeks-worth of tutoring sessions. 

Screenshot 2020-07-22 at 11 38 41 AM

samfalberg:
Here is the manage availability page with the calendar included. The calendar displays each session, with columns ordered by date. If there are multiple sessions on the same day, it will display all of them.
Screenshot 2020-07-23 at 10 29 05 AM

src/main/webapp/manage-availability.js
```diff
@@ -167,3 +167,54 @@ function deleteTimeSlot(window, timeslot) {
         }
     });
 }
+
+/** Creates a calendar with the Charts API and renders it on the page  */
+function createCalendar() {
+    fetch('/manage-availability', {method: 'GET'}).then(response => response.json()).then((timeslots) => {  
+        console.log(timeslots);
```
kashiish, line 174: Delete?

src/main/webapp/manage-availability.js
```diff
@@ -167,3 +167,54 @@ function deleteTimeSlot(window, timeslot) {
...
+        
+        const container = document.getElementById('calendar');
+        const chart = new google.visualization.Timeline(container);
+
+        const dataTable = new google.visualization.DataTable();
+        dataTable.addColumn({type: 'string', id: 'Date'});
+        dataTable.addColumn({type: 'date', id: 'Start'});
+        dataTable.addColumn({type: 'date', id: 'End'});
+        
+        for (var slot of timeslots) {
```
kashiish, line 191: Does the calendar show the availability for the week?

src/main/webapp/manage-sessions.js
```diff
@@ -144,3 +144,56 @@ function cancelTutorSession(window, scheduledSession) {
         }
     });
 }
+
+/** Creates a calendar with the Charts API and renders it on the page  */
+function createCalendar() {
+    fetch('/confirmation', {method: 'GET'}).then(response => response.json()).then((scheduledSessions) => {
+        console.log(scheduledSessions);
```
kashiish, line 151: Delete?

src/main/webapp/manage-sessions.js
```diff
@@ -144,3 +144,56 @@ function cancelTutorSession(window, scheduledSession) {
...
+        const container = document.getElementById('calendar');
+        const chart = new google.visualization.Timeline(container);
+
+        const dataTable = new google.visualization.DataTable();
+        dataTable.addColumn({type: 'string', id: 'Date'});
+        dataTable.addColumn({type: 'string', id: 'Description'});
+        dataTable.addColumn({type: 'date', id: 'Start'});
+        dataTable.addColumn({type: 'date', id: 'End'});
+        
+        for (var session of scheduledSessions) {
```
kashiish, line 169: Does this show all upcoming sessions or just for a specific time?

samfalberg: It shows all upcoming sessions

src/main/webapp/manage-availability.js
samfalberg: Right now it shows every available session, but we can update it so it only spans a single week.

kashiish: I think that's fine!

src/main/webapp/manage-availability.js
```diff
@@ -167,3 +167,53 @@ function deleteTimeSlot(window, timeslot) {
...
```
betrevisan, line 190: it might be better to create a separate function that is responsible for creating the calendar and calling inside this fetch. Maybe something like createCalendarHelper.

PR #88, opened 2020-07-22
Emails

betrevisan:
The commits proposed in this PR implement sending out emails. Changes include sending an email to the user once they first register, sending an email to both tutor and student when a tutoring session is scheduled, and sending an email to both tutor and student when a tutoring session is cancelled. Java tests were updated to check if the email was successfully sent.


src/main/java/com/google/sps/servlets/DeleteTutorSessionServlet.java
```diff
@@ -44,12 +55,92 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
         String studentID = Optional.ofNullable((String)request.getSession(false).getAttribute("userId")).orElse("-1");
         long id = Long.parseLong(request.getParameter("id"));
 
+        TutorSession tutorSession = datastore.getScheduledSession(id);
+
+        System.out.println(tutorSession.getStudentID());
```
kashiish, line 60: Delete?

src/main/java/com/google/sps/servlets/DeleteTutorSessionServlet.java
```diff
@@ -44,12 +55,92 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
         datastore.deleteTutorSession(studentID, id);
 
         String json = new Gson().toJson(datastore.getScheduledSession(id));
         response.setContentType("application/json;");
-        response.getWriter().println(json);
+        response.getWriter().println(json + "testTutorEmail: " + testTutorEmail + " testStudentEmail: " + testStudentEmail);
         return;
     }
+
+    public boolean sendCancellationEmailToTutor(Entity studentEntity, Entity tutorEntity, TimeRange timeslot) {
```
kashiish, line 78: If the only different between `sendCancellationEmailToTutor` and `sendCancellationEmailToStudent` is the message, maybe we can make `message` a parameter and use a single method. I think we only use the studentEntity in the tutor method for the student's name and the tutorEntity in the student method for the tutor's name. We could also do all of the date stuff outside of this method to create the message.

So we could have the method be `sendCancellationEmail(Entity entity, String message)`. Does that make sense?

src/main/java/com/google/sps/servlets/DeleteTutorSessionServlet.java
```diff
@@ -44,12 +55,92 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+        Properties props = new Properties();
+        Session session = Session.getDefaultInstance(props, null);
```
kashiish, line 80: What is a `Session` here?

src/main/java/com/google/sps/servlets/SchedulingServlet.java
```diff
@@ -71,8 +82,83 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+        boolean testStudentEmail = sendConfirmationEmailToStudent(studentEntity, tutorEntity, timeslot);
+
+
         String json = new Gson().toJson(datastore.getScheduledSessionsForTutor(tutorID));
-        response.getWriter().println(json);
+        response.getWriter().println(json + "testTutorEmail: " + testTutorEmail + " testStudentEmail: " + testStudentEmail);
         return;
     }
+
+    public boolean sendConfirmationEmailToTutor(Entity studentEntity, Entity tutorEntity, TimeRange timeslot) {
```
kashiish, line 97: We could do a similar thing here by consolidating these two methods into one?

src/main/java/com/google/sps/servlets/DeleteTutorSessionServlet.java
```diff
@@ -44,12 +55,92 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+        } catch (MessagingException e) {
+            System.out.println("Failed to send email.");
+            return false;
+        } catch (UnsupportedEncodingException e) {
+            System.out.println("Failed to encode email.");
+            return false;
+        }
+    }
+
+    public boolean sendCancellationEmailToStudent(Entity studentEntity, Entity tutorEntity, TimeRange timeslot) {
```
kashiish, line 112: Do these methods need to be public?

src/main/java/com/google/sps/servlets/SchedulingServlet.java
```diff
@@ -71,8 +81,41 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
 
+        sendConfirmationEmail("bernardo.trevisan@gmail.com");
+
         String json = new Gson().toJson(datastore.getScheduledSessionsForTutor(tutorID));
         response.getWriter().println(json);
         return;
     }
+
+    public void sendConfirmationEmail(String to) {
+        System.out.println("Create email");
```
samfalberg: No longer needed?

src/main/java/com/google/sps/servlets/DeleteTutorSessionServlet.java
betrevisan: Yes, this makes sense. I'll implement this for the cancellation and scheduling emails.

src/main/java/com/google/sps/servlets/DeleteTutorSessionServlet.java
betrevisan: It is basically the session accessed by the email host to create the message and send it out. Here's more info about that: https://stackoverflow.com/questions/22657918/what-is-the-use-of-javax-mail-session

src/main/java/com/google/sps/servlets/DeleteTutorSessionServlet.java
betrevisan: Not at all, I'll make them private. Thanks!

PR #91, opened 2020-07-23
Jasmine test fixes

kashiish:
These commits fix a few errors in the Jasmine tests for the profile and registration page. For the profile, the createProfileDiv call was missing a mock login status and there was a problem adding event listeners to certain elements. For the registration page, I updated the logout link test as we are not using the url provided by the Authentication API anymore.


src/main/webapp/registration.js
```diff
@@ -117,6 +113,15 @@ function displayLoginLogoutLinkHelper(document, loginStatus) {
...
 }
 
+/** Tells the server to log the user out and redirect to homepage. */
+function logout(window) {
+    fetch('/logout', {method: 'GET'}).then((response) => {
+        if(response.redirected) {
+            window.location.href = response.url;
+        }
+    });
+}
```
betrevisan, line 123: Good idea to separate the fetch function.

PR #96, opened 2020-07-23
Tasks for Rating Email

betrevisan:
The commits proposed in this PR implement the feature to send an email after a tutoring session has ended. Changes include adding the Google Cloud Tasks API, creating a task class for scheduling a task, and creating a servlet that sends the rating email once the time comes. The limitation of the Google Cloud Tasks API is that the maximum time in the future that tasks can be scheduled for is 720 hours.

I have also removed some of the unnecessary print statements I found on the way.


src/main/java/com/google/sps/data/RatingEmailTask.java
```diff
@@ -0,0 +1,93 @@
...
+        int hour = 10;
+        int min = 0;
+
+        Calendar scheduledDate = Calendar.getInstance();
+        scheduledDate.set(Calendar.YEAR, year);
+        scheduledDate.set(Calendar.MONTH, month);
+        scheduledDate.set(Calendar.DAY_OF_MONTH, day);
+        scheduledDate.set(Calendar.HOUR_OF_DAY, hour);
+        scheduledDate.set(Calendar.MINUTE, min);
+        this.scheduledSeconds = scheduledDate.getTime().getTime();
```
samfalberg, line 53: Why do you have to use the getTime method twice?

src/main/java/com/google/sps/servlets/SendRatingEmailServlet.java
```diff
@@ -64,7 +68,7 @@ public boolean sendRatingEmailToStudent(String email) {
         Session session = Session.getDefaultInstance(props, null);
 
         String subject = "Rate Your Tutoring Session";
-=        String message = "Hi there,\n" +
+        String message = "Hi there,\n" +
```
samfalberg, line 68: Is it possible to get the student's name and use it in this message? ('Hi [student name]' rather than 'Hi there')

src/main/java/com/google/sps/data/RatingEmailTask.java
betrevisan: The first one gets the Calendar in the format of Date and the second one translates date into milliseconds.

src/main/java/com/google/sps/data/RatingEmailTask.java
```diff
@@ -0,0 +1,98 @@
...
+                        AppEngineHttpRequest.newBuilder()
+                            .setBody(ByteString.copyFrom(payload, Charset.defaultCharset()))
+                            .setRelativeUri("/rating-email")
+                            .setHttpMethod(HttpMethod.POST)
+                            .build());
+
+            long current = Calendar.getInstance().getTime().getTime();
+
+            // Add the scheduled time to the request.
+            int seconds = (int) (this.scheduledSeconds - current) / 1000;
```
kashiish, line 85: Does this get the number of seconds until the session?

betrevisan: Yes. It is the number of milliseconds when the session ends minus the number of milliseconds from the time when the task is scheduled. I have to divide it by 1000 to get it into seconds, which is the format tasks asks for.

src/main/java/com/google/sps/servlets/SendRatingEmailServlet.java
betrevisan: Yes, I can work on doing this.

src/main/java/com/google/sps/data/RatingEmailTask.java
```diff
@@ -0,0 +1,98 @@
...
+            taskBuilder.setScheduleTime(
+                Timestamp.newBuilder()
+                    .setSeconds(Instant.now(Clock.systemUTC()).plusSeconds(seconds).getEpochSecond()));
```
kashiish, line 88: In that case, does this just set the seconds to the time of the scheduled session? Couldn't we just use scheduledSessions here?

betrevisan: The thing is that the end time in scheduledSessions is the number of milliseconds in total from january 1st 1970. However, what needs to be passed to the task builder is the seconds from the time when the tutoring session is scheduled. This is why I needed to subtract the current time to only get those seconds. If I just used the number of seconds from scheduled sessions it would be the equivalent of scheduling the task for 50 years from now.

PR #97, opened 2020-07-23
Profile fix

betrevisan:
The commits proposed in this PR have the goal of fixing issues with the display of the profile page. Changes include adjusting the way the profile.js file differentiates between a student and a tutor and passing down the student id to the history servlet when the person wanting to see the student's profile is that student's tutor.

### DIFF
--- a/src/main/java/com/google/sps/servlets/HistoryServlet.java
+++ b/src/main/java/com/google/sps/servlets/HistoryServlet.java
@@ -48,6 +48,11 @@ public class HistoryServlet extends HttpServlet {
         // Get the id of the student whose tutoring history will be displayed.
         // Make the default id -1 if the parameter is null, so an empty list will be returned
         String studentID = Optional.ofNullable((String)request.getSession(false).getAttribute("userId")).orElse("-1");
+        String studentIDTutorView = Optional.ofNullable(request.getParameter("studentIDTutorView")).orElse("-1");
+
+        if (!studentIDTutorView.equals("-1")) {
+            studentID = studentIDTutorView;
+        }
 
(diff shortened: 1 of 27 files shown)


src/main/java/com/google/sps/servlets/HistoryServlet.java
```diff
@@ -48,6 +48,11 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
         // Get the id of the student whose tutoring history will be displayed.
         // Make the default id -1 if the parameter is null, so an empty list will be returned
         String studentID = Optional.ofNullable((String)request.getSession(false).getAttribute("userId")).orElse("-1");
+        String studentIDTutorView = Optional.ofNullable(request.getParameter("studentIDTutorView")).orElse("-1");
```
kashiish, line 51: What is this used for?

src/main/webapp/profile.js
```diff
@@ -94,7 +94,7 @@ function fetchStatusHelper(user, loginStatus, window, document) {
     var text;
 
     // Display topics as comma-separated list with spaces
-    if (role == 'student') { 
+    if (user.learning != null) {
```
kashiish, line 97: Why aren't we using role here anymore?

src/main/java/com/google/sps/servlets/HistoryServlet.java
betrevisan: This is only passed down if the person trying to see the student's history is the student's tutor. If that is not the case, then the servlet will just load the history of the currently logged in user.

PR #102, opened 2020-07-24
Do not allow tutors to schedule sessions with themselves

kashiish:
These commits prevent tutors from seeing themselves in the search results. I have also added a check in SchedulingServlet to make sure that the tutorID and studentID are not the same in case someone tries to schedule directly from the servlet.


src/main/java/com/google/sps/servlets/SchedulingServlet.java
```diff
@@ -69,8 +69,14 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
         //if the tutor or student id was null
         if(tutorID.equals("-1") || studentID.equals("-1")) {
             response.getWriter().println("{\"error\": \"There was an error scheduling your session.\"}");
+            return;
         }
 
+        if(tutorID.equals(studentID)) {
+            response.getWriter().println("{\"error\": \"You cannot schedule your own session.\"}");
+            return;
+        }
```
betrevisan, line 78: Good check

PR #103, opened 2020-07-24
Search User by Name

betrevisan:
The commits proposed in this PR implement the ability to search users by their name (full name, first name, or last name). Changes include adding a search bar to the navigation bar, creating a file with js function for retrieving users and creating result boxes, creating java servlet to handle querying datastore for the users that match the given name, creating an user object class, updating name field properties of an user in datastore (inside the registration servlet), and writing the appropriate junit and jasmine tests.


src/main/java/com/google/sps/data/SampleData.java
```diff
@@ -93,6 +93,11 @@
                 new ArrayList<String> (Arrays.asList("0")), new ArrayList<TutorSession> (Arrays.asList()), "4")
```
kashiish, line 93: Now that we can search for users, we should add a link on tutors' profiles to their availability page so students can schedule session directly from there instead of searching from the homepage.

src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,40 @@
...
+
+package com.google.sps.data;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.LinkedList;
+
+/** Stores information about a user including name and userId. */
+public final class User {
```
kashiish, line 23: We can probably make the Student and Tutor class a subclass of the User class now.

kashiish: We can also probably add other common things like bio, pfp, email, etc to this class.

src/main/java/com/google/sps/utilities/SearchDatastoreService.java
```diff
@@ -76,6 +79,38 @@
...
+    public List<User> getUsersForName(String name) {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+        Filter fullNameFilter = new FilterPredicate("fullName", FilterOperator.EQUAL, name.toLowerCase());
+        Filter firstNameFilter = new FilterPredicate("firstName", FilterOperator.EQUAL, name.toLowerCase());
+        Filter LastNameFilter = new FilterPredicate("lastName", FilterOperator.EQUAL, name.toLowerCase());
+        CompositeFilter userFilter = CompositeFilterOperator.or(fullNameFilter, firstNameFilter, LastNameFilter);
+
+        Query userQuery = new Query("User").setFilter(userFilter);
+        //Query userQuery = new Query("User");
```
kashiish, line 95: Delete?

src/main/webapp/search-user-results.html
```diff
@@ -0,0 +1,83 @@
...
+                    <ul class="nav navbar-nav">
+                        <li><a href="#about">ABOUT</a></li>
+                        <li><a href="#about">CONTACT</a></li>
+                        <li id="account-dropdown" class="dropdown"><a class="dropdown-toggle" data-toggle="dropdown" href="#">MY ACCOUNT <span class="caret"></span></a>
+                            <ul class="dropdown-menu">
+                                <li><a id="profile">Profile</a></li>
+                                <li><a id="history">History</a></li>
+                                <li><a id="availability-settings">Manage Availability</a></li>
+                                <li><a id="my-students">My Students</a></li>
+                                <li><a id="my-progress">My Progress</a></li>
```
kashiish, line 43: Should we delete this?

src/main/webapp/search-users.js
```diff
@@ -0,0 +1,115 @@
...
+    container.appendChild(profileLink);
+
+    return container;
+}
+
+// Referenced to https://www.aspsnippets.com/Articles/Redirect-to-another-Page-on-Button-Click-using-JavaScript.aspx#:~:text=Redirecting%
+// 20on%20Button%20Click%20using%20JavaScript&text=Inside%20the%20Send%20JavaScript%20function,is%20redirected%20to%20the%20URL on June 23rd.
+// This function reads the id of the tutor that the student has selected and the start and end of the time range selected, which are
+// all passed as URI components, and adds them to the queryString array.
+function readComponents(queryString, window) {
```
kashiish, line 102: Is this the same function as the one in general.js?

src/main/java/com/google/sps/utilities/SearchDatastoreService.java
```diff
@@ -76,6 +79,38 @@
...
+    * Gets a list of users from datastore that have the specified name as a skill.
+    * @return List<User>
+    */
+    public List<User> getUsersForName(String name) {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+        Filter fullNameFilter = new FilterPredicate("fullName", FilterOperator.EQUAL, name.toLowerCase());
+        Filter firstNameFilter = new FilterPredicate("firstName", FilterOperator.EQUAL, name.toLowerCase());
+        Filter LastNameFilter = new FilterPredicate("lastName", FilterOperator.EQUAL, name.toLowerCase());
+        CompositeFilter userFilter = CompositeFilterOperator.or(fullNameFilter, firstNameFilter, LastNameFilter);
```
samfalberg, line 92: This is so cool, I didn't know you could do this

src/main/webapp/search-users.js
betrevisan: Yeah! Now that I merged general I will remove it :)

src/main/webapp/search-user-results.html
betrevisan: Yes, I just removed my-progress

src/main/java/com/google/sps/utilities/SearchDatastoreService.java
betrevisan: Thanks!

src/main/java/com/google/sps/data/User.java
betrevisan: Oh true! That would actually be super useful with the update to users being able to be both tutor and student. I'll do this in a new PR

src/main/java/com/google/sps/data/SampleData.java
betrevisan: good idea! I'll do this a new PR after this is merged then

PR #106, opened 2020-07-27
Add time and date inputs to enter availability

samfalberg:
These commits will add the HTML time and date input fields to the add-availability page. This will make entering availability easier.


src/main/webapp/manage-availability.js
```diff
@@ -130,13 +134,9 @@ function addTimeSlot() {
...
-    params.append('startHour', document.getElementById('startHour').value);
-    params.append('startMinute', document.getElementById('startMinute').value);
-    params.append('endHour', document.getElementById('endHour').value);
-    params.append('endMinute', document.getElementById('endMinute').value);
-    params.append('day', document.getElementById('day').value);
-    params.append('month', document.getElementById('month').value);
-    params.append('year', document.getElementById('year').value);
+    params.append('startTime', document.getElementById('startTime').value);
+    params.append('endTime', document.getElementById('endTime').value);
+    params.append('date', document.getElementById('date').value);
```
betrevisan, line 139: Is date a string or an object date? I guess my question is, to access month, year and day, can you do date.month or do you have to break down the string?

samfalberg: It's a string, so you'd have to split it to get the month, year, and day.

PR #107, opened 2020-07-27
Add rating and profile link to search results

kashiish:
These commits add the tutor's rating and a link to their profile when they show up in search results. I've also updated the tests for these new additions and fixed a few tests that were failing before.


src/main/webapp/search.js
```diff
@@ -269,31 +269,59 @@ function createBookResult(result) {
 function createTutorResult(result) {
     var container = document.createElement("div");
     var name = document.createElement("h3");
+    var rating = document.createElement("div");
     var email = document.createElement("h6");
     var skills = document.createElement("p");
     var availabilityLink = document.createElement("a");
 
-    name.innerText = result.name;
+    name.innerHTML = "<a style='color:black' href='/profile.html?userID=" + result.userId + "'>" + result.name + "</a>";
```
betrevisan, line 277: Cool

src/main/webapp/search.js
```diff
@@ -269,31 +269,59 @@ function createBookResult(result) {
...
     email.innerText = result.email;
     // Remove blank entry before adding to inner text
     var index = result.skills.indexOf(' ');
     result.skills.splice(index, 1);
     skills.innerText = "Skills: " + result.skills.join(", ");
     availabilityLink.innerText = "Availability";
+    loadStars(rating, Math.round(parseInt(result.ratingSum) / parseInt(result.ratingCount)));
```
betrevisan, line 284: Wouldn't rating sum always be an integer? Users can only rate tutors using integers so I think the sum will always be an integer.

kashiish: I kept getting NaN when dividing so I thought ratingSum and ratingCount were Strings but it was just dividing by 0. Forgot to delete parseInt!

PR #111, opened 2020-07-27
Allow users to sort by rating

kashiish:
These commits add a filter list in the search results so users can sort by rating or names (alphabetically). I added a new property in the Tutor entity called "rating" that will keep track of the tutor's rating and allow us to get data from datastore that is already sorted by rating. I also added new tests for this feature and created a method in SampleData that will allow us to rate a tutor for testing.


src/main/java/com/google/sps/data/SampleData.java
```diff
@@ -162,6 +174,9 @@ public void addStudentsToDatastore() {
         }
     }
 
+    /**
+    * Adds all sample User objects to datastore.
+    */
```
betrevisan, line 179: Thanks for adding the comments in this file

PR #112, opened 2020-07-27
Update calendar and availability 

samfalberg:
These commits will add update the calendar display and functionality. Changes will include adding a calendar to the availability page, automatically making tutor sessions 1 hour long, adding tutor names to rows on the manage sessions page, making the calendar display dates in chronological order, and adding event listeners to sessions so you can delete and schedule sessions directly from the calendar.

### DIFF
--- a/src/main/webapp/availability.js
+++ b/src/main/webapp/availability.js
@@ -56,8 +56,9 @@ function createTimeSlotBox(timeslot, tutorID) {
         amOrPm = "pm"
     }
     var minute = parseInt(timeslot.start) % 60;
-    if (minute == 0) {
-        minute = "00";
+    // Display minutes regularly (e.g. 1:05pm rather than 1:5pm)
+    if (minute < 10) {
+        minute = "0" + minute;
     }
     dateElement.innerText = hour + ":" + minute + amOrPm + " on " + months[timeslot.date.month] + " " + timeslot.date.dayOfMonth + ", " + timeslot.date.year;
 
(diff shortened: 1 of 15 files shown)


src/main/webapp/availability.js
```diff
@@ -12,86 +12,103 @@
...
-            timeslots.forEach((timeslot) => {
-                document.getElementById('timeslots').appendChild(createTimeSlotBox(timeslot, tutorID));
-            });
-        } else {
-            var timeslotsContainer = document.getElementById('timeslots');
-            var errorMessage = document.createElement("p");
-            errorMessage.innerText = "This user has not set any available timeslots";
-            timeslotsContainer.appendChild(errorMessage);
+        // Don't create a calendar if there are no available timeslots
+        if (timeslots === undefined || timeslots.length == 0) {
```
betrevisan, line 30: Could you keep the error message that is displayed if there are no timeslots?

src/main/webapp/confirmation.js
```diff
@@ -12,86 +12,95 @@
...
-function getScheduledSessions() {
-    return getScheduledSessionsHelper(window);
-}
-
-//Helper function for getScheduledSessions, used for testing
-async function getScheduledSessionsHelper(window) {
-    await fetch('/confirmation', {method: 'GET'}).then((response) => {
+/** Creates a calendar with the Charts API and renders it on the page  */
+function createCalendar() {
+    fetch('/confirmation', {method: 'GET'}).then((response) => {
```
betrevisan, line 17: Shouldn't the await still be here?

src/main/webapp/confirmation.js
```diff
@@ -12,86 +12,95 @@
...
-
-function createScheduledSessionBox(scheduledSession) {
-
-    var months = [ "January", "February", "March", "April", "May", "June", 
-           "July", "August", "September", "October", "November", "December" ];
-
-    const scheduledSessionElement = document.createElement('li');
-    scheduledSessionElement.className = 'list-group-item';
+        
+        const chart = new google.visualization.Timeline(container);
```
betrevisan, line 43: Maybe add the part that constructs the calendar as a separate function from the part that fetches confirmation

src/main/webapp/availability.js
```diff
@@ -12,86 +12,103 @@
...
 
-    const selectButtonElement = document.createElement('button');
-    selectButtonElement.innerText = 'Select';
-    selectButtonElement.style.textAlign = 'right';
-    selectButtonElement.style.display = 'inline';
-    selectButtonElement.className = 'btn btn-default btn-lg';
-    selectButtonElement.addEventListener('click', () => {
-        selectTimeSlot(tutorID, window, timeslot);
-    });
+        // Have timeline span 24 hours regardless of what's scheduled
```
kashiish, line 58: Does this show the entire day now instead of just the scheduled session?

src/main/webapp/manage-availability.js
```diff
@@ -133,43 +43,45 @@ function addTimeSlotHelper(window) {
...
+    fetch('/manage-availability', {method: 'GET'}).then((response) => {
+        //if the student is not the current user or not signed in
         if(response.redirected) {
             window.location.href = response.url;
-            alert("You must be signed in to edit availability.");
+            alert("You must be signed in to manage availability.");
+            return [];
+        }
+        return response.json();
+    }).then((timeslots) => {
```
betrevisan, line 56: Keep the message that used to be displayed if the user does not have any available timeslots to manage

src/main/webapp/manage-availability.js
```diff
@@ -133,43 +43,45 @@ function addTimeSlotHelper(window) {
...
 
         const dataTable = new google.visualization.DataTable();
         dataTable.addColumn({type: 'string', id: 'Date'});
         dataTable.addColumn({type: 'date', id: 'Start'});
         dataTable.addColumn({type: 'date', id: 'End'});
+
+        // Sort timeslots in chronological order
+        timeslots.sort(function(a, b) {
+            return (a.date.year + a.date.month/12 + a.date.dayOfMonth/365) - (b.date.year + b.date.month/12 + b.date.dayOfMonth/365);
+        });
```
betrevisan, line 84: Separate this and the below parts into a separate function

src/test/java/com/google/sps/ManageAvailabilityTest.java
```diff
@@ -120,7 +119,6 @@ public void testDoPost() throws Exception {
         servlet.doPost(request, response);
 
         verify(request, times(1)).getParameter("startTime");
-        verify(request, times(1)).getParameter("endTime");
         verify(request, times(1)).getParameter("date");
 
         Calendar expectedDate = new Calendar.Builder()
```
kashiish, line 124: I might have missed it, but are we adding the checks to make sure the tutor doesn't add an availability in the past?

src/main/webapp/availability.js
samfalberg: Sure. I forgot to add it earlier.

src/main/webapp/confirmation.js
samfalberg: It seemed to be working without it, but I added it back just in case.

src/main/webapp/availability.js
samfalberg: Yup

PR #113, opened 2020-07-28
Implement Blobstore

samfalberg:
These commits will use blobstore to allow the user to submit a custom profile picture. Since this feature is optional, if the user does not submit anything then their profile picture will be displayed as a hard-coded default image.


src/main/webapp/registration.html
```diff
@@ -75,7 +75,7 @@
       
         <div class="bg-description-blue">
             <div class="container-fluid bg-white">
-                <form id="registration-form" action="/registration" method="POST" style="display:none">
+                <form id="registration-form" enctype="multipart/form-data" method="POST" style="display:none">
```
kashiish, line 78: What does this do?

src/test/java/com/google/sps/BlobUploadUrlTest.java
```diff
@@ -0,0 +1,59 @@
...
+        PrintWriter writer = new PrintWriter(stringWriter);
+        when(response.getWriter()).thenReturn(writer);
+
+        servlet.setResponse(request, response, blobstore);
+
+        String expected = "/registration.html";
+        Assert.assertTrue(stringWriter.toString().contains(expected));
+    }
+
+}
```
kashiish, line 59: New line here

src/main/java/com/google/sps/servlets/RegistrationServlet.java
```diff
@@ -146,6 +158,29 @@ public void createUserEntityAndPutInDatastore(DatastoreService ds, Entity entity
...
+    Map<String, List<BlobKey>> blobs = blobstoreService.getUploads(request);
+    List<BlobKey> blobKeys = blobs.get(name);
+
+    // User submitted form without selecting a file, so we can't get a URL. (dev server)
+    if (blobKeys == null || blobKeys.isEmpty()) {
+        return null;
+    } 
+
+    // Form only contains a single file input, so get the first index.
+    String blobKey = blobKeys.get(0).getKeyString();
```
kashiish, line 172: Don't know if this really matters, but should we put this line after we check if the user submitted the form without a pfp?

src/main/webapp/registration.html
samfalberg: multipart/form-data is used for encoding data that involves a file input. Here's a good thread explaining it in more detail
https://stackoverflow.com/questions/4526273/what-does-enctype-multipart-form-data-mean

src/main/java/com/google/sps/servlets/BlobUploadUrlServlet.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+@WebServlet("/blob-upload-url")
+public class BlobUploadUrlServlet extends HttpServlet {
+
+    private BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+
+    @Override
+    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        setResponse(request, response, blobstoreService);
+    }
```
betrevisan, line 37: What is the use of this doGet? What does setResponse do?

src/main/java/com/google/sps/servlets/BlobUploadUrlServlet.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public void setResponse(HttpServletRequest request, HttpServletResponse response, BlobstoreService blobstoreService) throws IOException {
+        String uploadUrl = blobstoreService.createUploadUrl("/registration");
```
betrevisan, line 40: Does this mean the user will always be redirected to registration.html when uploading a picture?

src/main/java/com/google/sps/servlets/BlobUploadUrlServlet.java
samfalberg: setResponse is just a helper function for testing. I forgot to add a comment for it

src/main/java/com/google/sps/servlets/RegistrationServlet.java
samfalberg: It doesn't matter, but it makes more sense to have it afterwards yeah

PR #115, opened 2020-07-28
Allow users to sort by other properties

kashiish:
These commits create a select input where users can choose to sort tutors by name (alphabetical), rating, number of available time slots, number of students in ascending order, and number of students in descending order. I replaced the radio button sort by form to a select form because there are more options now. I also wrote JUnit tests for these sort filters and updated the Jasmine test to use a select form.


src/main/java/com/google/sps/data/SampleData.java
```diff
@@ -89,9 +89,9 @@
         new Student("Kashish Arora", "Kashish\'s bio", "images/pfp.jpg", "kashisharora@google.com", new ArrayList<String> (Arrays.asList("English", "Physics")),
                 new ArrayList<String> (Arrays.asList()), new ArrayList<TutorSession> (Arrays.asList()), "0"),
         new Student("Bernardo Eilert Trevisan", "Bernardo\'s bio", "images/pfp.jpg", "btrevisan@google.com", new ArrayList<String> (Arrays.asList("Math", "History")),
-                new ArrayList<String> (Arrays.asList()), new ArrayList<TutorSession> (Arrays.asList(bernardoSession)), "1"),
+                new ArrayList<String> (Arrays.asList("1")), new ArrayList<TutorSession> (Arrays.asList(bernardoSession)), "1"),
```
betrevisan, line 92: What does the number parameter in asList() mean? Is it also the id?

src/main/java/com/google/sps/utilities/SearchDatastoreService.java
```diff
@@ -90,10 +96,61 @@
...
+
+        Collections.sort(tutors, new Comparator<Tutor>() {
+            @Override
+            public int compare(Tutor a, Tutor b) {
+                if(order.equals("most")) {
+                    return b.getStudents().size() - a.getStudents().size();
+                }
+
+                return a.getStudents().size() - b.getStudents().size();
+            }
```
betrevisan, line 137: I don't know if got to look at this, but it could be cool to see what kind of sorting algorithms might be more efficient when making this comparison

src/main/java/com/google/sps/data/SampleData.java
kashiish: That arraylist contains the ids of the student's tutors.

src/main/java/com/google/sps/utilities/SearchDatastoreService.java
kashiish: Yeah that could be cool. I think Collections.sort uses merge sort which is pretty good for longer lists.

PR #120, opened 2020-07-30
Update web.xml

betrevisan:
The commit proposed in this PR adds the new servlets related to the groups feature to the web.xml in order to prevent users from posting, replying, or creating groups if they are not logged in,

### DIFF
--- a/src/main/webapp/WEB-INF/web.xml
+++ b/src/main/webapp/WEB-INF/web.xml
@@ -18,7 +18,10 @@
         <url-pattern>delete-tutor-session</url-pattern>
         <url-pattern>history</url-pattern>
         <url-pattern>manage-availability</url-pattern>
+        <url-pattern>manage-groups</url-pattern>
         <url-pattern>manage-lists</url-pattern>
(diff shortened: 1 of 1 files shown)


src/main/webapp/WEB-INF/web.xml
```diff
@@ -18,7 +18,10 @@
         <url-pattern>delete-tutor-session</url-pattern>
         <url-pattern>history</url-pattern>
         <url-pattern>manage-availability</url-pattern>
+        <url-pattern>manage-groups</url-pattern>
```
kashiish, line 21: Because manage-groups/replies/posts also GET groups/posts/replies, viewers that are not logged in will not be able to view group stuff. Should we allow not logged in users to view groups?

betrevisan: Good point. I think it might be good in terms of privacy to make sure that only logged in users can see the posts and also in terms of getting more users (since people would have more reasons to register). What do you think?

kashiish: Yeah, that makes sense.

PR #123, opened 2020-07-30
Delete available timeslots that bleed into next day

samfalberg:
These commits will fix the issue where a tutor can't delete a timeslot that bleeds into the next day (i.e. 11:30pm - 12:30am). Changes include added logic in AvailabilityDatastoreService.java, and deletion of an unnecessary doGet function.

### DIFF
--- a/src/main/java/com/google/sps/servlets/DeleteAvailabilityServlet.java
+++ b/src/main/java/com/google/sps/servlets/DeleteAvailabilityServlet.java
@@ -40,12 +40,6 @@ public class DeleteAvailabilityServlet extends HttpServlet {
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/servlets/DeleteAvailabilityServlet.java
```diff
@@ -39,12 +39,6 @@ public void init() {
         datastore = new AvailabilityDatastoreService();
     }
 
-    @Override
-    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-        response.setContentType("plain/text");
-        response.getWriter().println("To be implemented");
-    }
```
betrevisan, line 46: Thanks for noticing this

PR #125, opened 2020-07-30
Fetch Google Books data through servlet

kashiish:
These commits create a new servlet called GoogleBooksServlet that fetches books data with the parameters passed from JavaScript fetch requests. I updated the JavaScript fetch calls from the API to the new servlet, so the API key is no longer exposed in the front end. Instead I stored the key in an environment variable in Cloud Shell. 

I also added a Jasmine test for the load more books function. 

I will update mylists.js after my other PR is merged so that it makes a request to the servlet instead of the API directly. I will also work on the servlet test later as it will take me some time to figure out how exactly to test it. We needed to hide the API key ASAP, so I thought it would be better to submit a PR without a test for now.

### DIFF
--- a/src/main/java/com/google/sps/servlets/GoogleBooksServlet.java
+++ b/src/main/java/com/google/sps/servlets/GoogleBooksServlet.java
@@ -18,9 +18,7 @@ import com.google.gson.Gson;
 import com.google.gson.JsonParser;
 import com.google.gson.JsonObject;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
+import java.util.Optional;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/servlets/GoogleBooksServlet.java
```diff
@@ -0,0 +1,84 @@
...
+
+        if(author != null) {
+            params += "+inauthor:" + author;
+        }
+
+        if(startIndex >= 0) {
+            params += "&startIndex=" + startIndex;
+        }
+
+        URL url = new URL("https://www.googleapis.com/books/v1/volumes?" + params + "&key=" + System.getenv("GOOGLE_API_KEY"));
```
betrevisan, line 65: This looks good

src/main/java/com/google/sps/servlets/GoogleBooksServlet.java
```diff
@@ -0,0 +1,84 @@
...
+        HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
betrevisan, line 66: What is a connection exactly?

src/main/java/com/google/sps/servlets/GoogleBooksServlet.java
```diff
@@ -0,0 +1,84 @@
...
+        String line;
+        StringBuffer output = new StringBuffer();
+
+        while ((line = stream.readLine()) != null) {
+            output.append(line);
+        }
+
+        stream.close();
+
+        JsonObject jsonObject = new JsonParser().parse(output.toString()).getAsJsonObject();
```
betrevisan, line 79: Can't gson be used here?

src/main/java/com/google/sps/servlets/GoogleBooksServlet.java
kashiish: It's kind of like `fetch` in Javascript. It allows us to make a request to a URL.

src/main/java/com/google/sps/servlets/GoogleBooksServlet.java
kashiish: I don't think you can use .toJson() on Strings. JsonParser and JsonObject are part of the Gson library.

PR #128, opened 2020-07-30
Prohibit scheduling past times

samfalberg:
These commits will fix the issue where the user is not able to input a past time when they are trying to schedule a session for a future day. Changes include removing the min attribute from the time input, and adding a function that shows an alert when the time is in the past.


src/main/webapp/manage-availability.js
```diff
@@ -43,35 +52,49 @@ function addTimeSlotHelper(window) {
...
+    }
+    if (day < 10) {
+        day = "0" + day;
+    }
+    var yearMonthDay = year + '-' + month + '-' + day;
+    var startTimeArr = startTime.split(':');
+
+    // Add check to see if user-inputted time is in the past
+    if (yearMonthDay == sessionDate) {
+        if (Number(startTimeArr[0]*60) + Number(startTimeArr[1]) < hour*60 + minute) {
```
betrevisan, line 69: Why is this number? Isn't it better to parse it into a long or an int?

samfalberg: I changed it to parseInt

PR #129, opened 2020-07-30
Add Group Header

betrevisan:
The commits proposed int his PR load the group header when that group's page is viewed. Changes include adding a servlet that gets information about a group given its id and adding a loadGroupHeader function to group.js. Java and Jasmine tests have also been added

### DIFF
--- a/src/test/java/com/google/sps/GroupDataTest.java
+++ b/src/test/java/com/google/sps/GroupDataTest.java
@@ -1,0 +1,155 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/sps/GroupDataTest.java
```diff
@@ -0,0 +1,155 @@
...
+        servlet.doGet(request, response);
+
+        verify(request, times(1)).getParameter("groupId");
+
+        String expected = new Gson()
+                            .toJson(new Group("test", "math", "", "123", 1));
+      
+        writer.flush();
+
+        // The correct list of experiences should have been returned
```
kashiish, line 128: Update comment here

PR #136, opened 2020-07-31
Fix logout issue on deployed version

kashiish:
These commits the logout issue we were having after we restricted the app to internal users only. Now, a user cannot log out but can switch accounts only as logging out is not possible with the IAP. When a user logs in to the app for the first time, they will still be redirected to the registration page. If a user logs in and they have not registered before, they will be redirected to the registration page.

This PR also fixes a few Jasmine tests and adds a test to check if the app redirects correctly when a user needs to register.


src/main/webapp/group.js
```diff
@@ -22,17 +22,17 @@ function addEventListeners() {
...
-function loadGroupHeaderHelper(document, window) {
+async function loadGroupHeaderHelper(document, window) {
     const params = new URLSearchParams();
 
     var queryString = new Array();
     window.onload = readComponents(queryString, window);
     const groupId = queryString["groupId"];
 
-    fetch("/group-data?groupId=" + groupId).then(response => response.json()).then((result) => {
+    await fetch("/group-data?groupId=" + groupId).then(response => response.json()).then((result) => {
```
betrevisan, line 35: Why is there a need to await now?

kashiish: This is mainly for the tests. The container wasn't being updated in the tests so I had to add an await to make sure the fetch call finished before I tested it.

PR #137, opened 2020-07-31
Hide availability on profiles for tutors

samfalberg:
These commits fix the issue where tutors can see their own availability, or other tutors' availability, when looking at tutor profiles. This should only be visible to students.


src/main/webapp/profile.js
```diff
@@ -133,10 +133,15 @@ function fetchStatusHelper(user, loginStatus, window, document) {
         text = "I am tutoring in: " + listWithSpaces;
         getListsProfile();
 
-        document.getElementById('tutor-availability').style.display = 'block';
-        document.getElementById('tutor-availability-btn').addEventListener('click', () => {
-            redirectToAvailability(user, window); 
-        });
+        // Only students should be allowed to schedule sessions with tutors from their profile
+        if (role == 'student') {
```
betrevisan, line 137: Either add || role == "both" or create a new conditional that handles the case in which a user is a student and also a tutor.

PR #140, opened 2020-07-31
Fix Groups nav bar issue and update Jasmine tests

kashiish:
These commits fix the id issue for the groups.html page. I've also updated a few Jasmine tests for the groups and group pages that I missed before.

### DIFF
--- a/src/main/webapp/group.js
+++ b/src/main/webapp/group.js
@@ -106,8 +106,6 @@ async function getPosts(groupId, document, window) {
         var postsContainer = document.getElementById("posts");
 
         var numSearchResults = document.getElementById("num-posts-results");
-
-        postsContainer.appendChild(numSearchResults);
         
         //if there was an error reported by the servlet, display the error message
         if(results.error) {
--- a/src/main/webapp/groups.js
+++ b/src/main/webapp/groups.js
@@ -49,11 +49,9 @@ async function getSearchGroupResultsHelper(document, window) {
 /** Fetches the list of groups for the group name the user searched for. */
 async function getGroups(group) {
     await fetch("/manage-groups?group=" + group).then(response => response.json()).then((results) => {
-        var groupContainer = document.getElementById("groups-results");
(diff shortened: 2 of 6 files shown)


src/main/webapp/groups.js
```diff
@@ -52,7 +52,7 @@ async function getGroups(group) {
         var groupContainer = document.getElementById("groups-result");
 
         var numSearchResults = document.getElementById("num-groups-results");
-        
+        console.log(numSearchResults);
```
samfalberg, line 55: Delete?

src/main/webapp/group.js
```diff
@@ -106,8 +106,6 @@ async function getPosts(groupId, document, window) {
         var postsContainer = document.getElementById("posts");
```
betrevisan, line 106: I think the id for this one should be posts-results

src/main/webapp/group.js
```diff
@@ -106,8 +106,6 @@ async function getPosts(groupId, document, window) {
...
 
         var numSearchResults = document.getElementById("num-posts-results");
-
-        postsContainer.appendChild(numSearchResults);
```
betrevisan, line 110: Shouldn't this be kept?

src/main/webapp/search-users.js
```diff
@@ -54,8 +54,6 @@ async function getUsers(name) {
         var userContainer = document.getElementById("users");
 
         var numSearchResults = document.getElementById("num-users-results");
-
-        userContainer.appendChild(numSearchResults);
```
betrevisan, line 58: Shouldn't this one also be kept?

src/main/webapp/group.js
kashiish: I added the results label H4 element to the HTML directly, so we can just get it by calling getElementById.

src/main/webapp/search-users.js
kashiish: Same thing here.

PR #141, opened 2020-07-31
Minutes Formatting in History

betrevisan:
The commits proposed inside this PR make sure that the formatting for minutes is correct inside the tutoring session boxes of history.html. The previous issue was that when the minutes of a time slot was between 1 and 9 it would be displayed as, for example, 9:2 am. The updates should fix this. I also added some jasmine tests to make sure this formatting is maintained.


src/main/webapp/history.js
```diff
@@ -82,6 +82,8 @@ function createTutoringSessionBox(tutoringSession) {
     var minute = parseInt(tutoringSession.timeslot.start) % 60;
     if (minute == 0) {
         minute = "00";
+    } else if (minute < 10) {
+        minute = "0" + minute;
```
samfalberg, line 86: I think you can get rid of the if statement and just keep the else if and it will work the same.

betrevisan: True! Thanks

PR #142, opened 2020-07-31
Store API key in .properties file and get rid of API key from front-end

kashiish:
These commits update the GoogleBooksServlet to get the GCP API key from a .properties file in the resources folder. I also updated the my-lists and search pages to make a request to the servlet instead of the API directly, so the key is no longer exposed. Lastly, I updated the Jasmine tests to reflect this change.

I will create a test for the GoogleBooksServlet in a separate PR.


src/main/java/com/google/sps/servlets/GoogleBooksServlet.java
```diff
@@ -47,24 +51,34 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
         }
 
-        URL url = new URL("https://www.googleapis.com/books/v1/volumes?" + params + "&key=" + System.getenv("GOOGLE_API_KEY"));
+        //get .properties file that stores the api key
+        Properties prop = new Properties();
+        InputStream input = getClass().getResourceAsStream("/api.properties");
+        prop.load(input);
+
+        //need country param here for appengine copyright purposes
+        URL url = new URL("https://www.googleapis.com/books/v1/volumes?" + params + "&maxResults=40&country=US");
```
betrevisan, line 75: This was super interesting :)

src/main/resources/.gitignore
```diff
@@ -0,0 +1 @@
+api.properties
```
betrevisan, line 1: Do we need to commit this to the repo?

src/main/java/com/google/sps/servlets/GoogleBooksServlet.java
kashiish: App engine is crazy

src/main/resources/.gitignore
kashiish: I thought it would be good to commit the folder just in case we need to add more resource files. The gitignore is there so the api.properties file (which stores the key) doesn't get uploaded.

PR #146, opened 2020-08-03
Add test for GoogleBooksServlet

kashiish:
These commits create a test class for the GoogleBooksServlet class. I separated the doGet method in the servlet into smaller helper methods so we could spy on the helper methods (like we did in the Jasmine tests).

### DIFF
--- a/src/main/java/com/google/sps/servlets/GoogleBooksServlet.java
+++ b/src/main/java/com/google/sps/servlets/GoogleBooksServlet.java
@@ -30,7 +30,6 @@ import java.io.InputStreamReader;
 import java.util.Properties;
 import java.io.InputStream;
 import java.net.URLEncoder;
-import java.io.File;
 
 /** Fetches data from the Google Books API. */
 @WebServlet("/books")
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/sps/servlets/GoogleBooksServlet.java
```diff
@@ -66,21 +65,50 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
         }
 
-        //get .properties file that stores the api key
-        Properties prop = new Properties();
-        InputStream input = getClass().getResourceAsStream("/api.properties");
-        prop.load(input);
+        HttpURLConnection connection = getConnection(params);
+        connection.setRequestMethod("GET");
+        connection.setRequestProperty("key", getApiKey());
+        connection.setRequestProperty("User-Agent", "Mozilla/4.76");
```
betrevisan, line 71: Could you explain what this User-Agent property does?

kashiish: It lets the server know what type of application is being used. In this case, I set it to use the Mozilla web browser.

